PR #2, opened 2023-02-20
[SIL-51] Add basic dependecies

Krzysiudan:
https://silentus-au.atlassian.net/jira/software/projects/SIL/boards/1?selectedIssue=SIL-51

Changes: 
- added DI Hilt dependency 
- added coroutines
- added retrofit
- added ktlint
- updated other dependencies.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,6 +1,8 @@
 plugins {
+    id 'kotlin-kapt'
     id 'com.android.application'
     id 'org.jetbrains.kotlin.android'
+    id 'com.google.dagger.hilt.android'
 }
 
 android {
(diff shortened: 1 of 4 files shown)


app/build.gradle
```diff
@@ -47,17 +50,39 @@ android {
...
-    implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.3.1'
-    implementation 'androidx.activity:activity-compose:1.3.1'
+    //Compose
     implementation "androidx.compose.ui:ui:$compose_version"
     implementation "androidx.compose.ui:ui-tooling-preview:$compose_version"
-    implementation 'androidx.compose.material3:material3:1.0.0-alpha11'
+    implementation 'androidx.compose.material3:material3:1.1.0-alpha06'
+
+    //DI - Hilt
+    implementation "com.google.dagger:hilt-android:2.45"
```
kezc, line 74: nitpick: if we extract version of compose maybe we could extract version of hilt and coroutines too

app/src/main/java/com/appunite/loudius/LoudiusApplication.kt
```diff
@@ -0,0 +1,9 @@
+package com.appunite.loudius
+
+import android.app.Application
+import dagger.hilt.android.HiltAndroidApp
+
+@HiltAndroidApp
+class LoudiusApplication : Application() {
```
kezc: nitpick: unnecessary parentheses

kezc (approved):

app/build.gradle
Krzysiudan: Hmm, why not.. 
Then we agree to extract every dependency version? Maybe we could use a gradle kotlin DSL?

nowakweronika (approved):
lgtm

PR #5, opened 2023-02-20
[SIL-59] Add linter to the CI

Krzysiudan:
https://silentus-au.atlassian.net/jira/software/projects/SIL/boards/1/backlog?selectedIssue=SIL-42

Changes: 
- configured MEGA-LINTER CI 😓 

There were a lot of troubles with configuring this action - probably too much. I've disabled one linter with spelling and linter for groovy because it had weird/stupid errors(IMHO). 

Proposition: 
We can use it for a while and delete it when/if we will get annoyed with the behaviour of that GitHub action. It is probably more suitable for a big project where it is beneficial to have everything checked. Anyway, I believe it's worth trying and then we can leave our opinions here: 
https://www.notion.so/appunite/Mega-linter-github-action-83fc41607d25412a9a53c0c7b2c2cc8a

If you agree to have that for a while, we have 2 options of auto-correcting small errors:
- auto commits 
- auto PR's 

PS: I will clean the git history after CR 😅


.github/workflows/run-code-quality-check.yml
```diff
@@ -0,0 +1,84 @@
+---
+# MegaLinter GitHub Action configuration file
+# More info at https://megalinter.github.io
+name: MegaLinter
+
+on:
+  push:
+  pull_request:
+    branches: [ master,main,develop ]
```
kezc: nitpick: we don't need master :P

github_conf/branch_protection_rules.json
```diff
@@ -0,0 +1,4 @@
+{
```
kezc, line 1: question: what is it?

.github/workflows/run-code-quality-check.yml
```diff
@@ -0,0 +1,84 @@
+---
```
kezc, line 1: In my opinion, it's better when the autoformatting is done in separate PR so we can choose if we want to apply it. However, do you know if it is possible that the new PR would be linked inside the original PR? I mean, I'd like to know that there's a PR related to my PR, because the title `[[MegaLinter] Apply linters automatic fixes]` doesn't say much and if we send the link on slack it might me omitted too :)

Krzysiudan: It's definitely possible to rename it and link it in the comments. I bet it will require a few additional hours, to find a proper reporter, take the link somehow and make it work. So I'm not sure it is worth it. 

But do we really have a choice if we want to fix something or not? If we decide not to then the linter will fail anyway. So probably these are needed anyway. We can try one option for now and then switch to the other later to see this solution in practice. Right now I would go for auto commits because I don't want to dig into that for a while 😅

github_conf/branch_protection_rules.json
Krzysiudan: It's an auto-generated file needed for that actions to work. It's a GitHub configuration file. I'm not so sure for what, but it won't work without that. I've tried to find out the reasons but without success.

.github/workflows/run-code-quality-check.yml
kezc: 1. That makes sense, I thought that maybe you know some easy way or something :)
2. Okay, if you need to merge it anyway then I think auto commits will fit better

kezc (approved):

nowakweronika (approved):

PR #6, opened 2023-02-21
[SIL-64] Api handling

nowakweronika:


Krzysiudan (approved):
**Suggestion**: Maybe let's put the ApiCallUtil.kt file with the safeApiCall() method in this PR too.

app/src/main/java/com/appunite/loudius/network/WebException.kt
```diff
@@ -0,0 +1,17 @@
+package com.appunite.loudius.network
+
+import java.io.IOException
+
+sealed class WebException : Exception() {
```
kezc, line 5: thought: Maybe we could include e.g. `safeApiCall` now? I think it would be helpful, but it will require some changes, as it was adjusted to crazy silentus API :D

kezc (approved):

app/src/main/java/com/appunite/loudius/network/ApiCallUtil.kt
```diff
@@ -0,0 +1,42 @@
+package com.appunite.loudius.network
+
+import retrofit2.HttpException
+import retrofit2.Response
+import java.io.IOException
+
+suspend fun <T : StatusTracker> safeApiCall(
```
Krzysiudan: **issue**: StatusTracker interface almost certainly won't be needed. 
So following function name would look differently: 
`suspend fun <T> safeApiCall(`
And below function same without extending `StatusTracker`

app/src/main/java/com/appunite/loudius/network/ApiCallUtil.kt
```diff
@@ -0,0 +1,42 @@
...
+        val status = body.status
+        if (status == 200) {
+            Result.success(body)
+        } else {
+            Result.failure(errorParser(status, body.message))
+        }
+    } else {
+        Result.failure(errorParser(response.code(), response.message()))
+    }
+}
```
Krzysiudan: **issue**: `StatusTracker` interface is not needed. 
Thanks to that we can remove extra checking is a successful response have the additional internal status of 200. So the method could look like this:  
```
private suspend fun <T> handleSuccessfulCall(
    apiCall: suspend () -> Response<T>,
    errorParser: RequestErrorParser
): Result<T> {
    val response = apiCall()
    val body = response.body()
    return if (response.isSuccessful && body != null) {
        Result.success(body)
    } else {
        Result.failure(errorParser(response.code(), response.message()))
    }
}
```

app/src/main/java/com/appunite/loudius/network/StatusTracker.kt
```diff
@@ -0,0 +1,6 @@
+package com.appunite.loudius.network
+
+interface StatusTracker {
+    val status: Int
+    val message: String
+}
```
Krzysiudan: **issue** Not needed interface. 
Let's remove that bastard. 😅

Krzysiudan:
**nitpick**: Autoformatting code says "I want to organize your imports too" 😂 : 
Screenshot 2023-02-21 at 11 29 32

app/src/main/java/com/appunite/loudius/network/ApiCallUtil.kt
```diff
@@ -0,0 +1,37 @@
...
+
+private suspend fun <T> handleSuccessfulCall(
+    apiCall: suspend () -> Response<T>,
+    errorParser: RequestErrorParser
+): Result<T> {
+    val response = apiCall()
+    val body = response.body()
+    return if (response.isSuccessful && body != null) {
```
kezc: suggestion: Do we need to check the response by hand if we don't have the StatusTracker? Can't the service just return the object itself without a wrapper? we would need to catch HttpException instead of the if, but I think it would make the code a little bit cleaner. WDYT? Also, I'm not sure about checking body != null, what if there's no body, because it's e.g. DELETE method? https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#delete-a-repository

nowakweronika: You're right. Even if it's not DELETE method, we don't need to check the response :)

PR #9, opened 2023-02-23
[SIL-65] Details screen UI

Krzysiudan:
https://silentus-au.atlassian.net/jira/software/projects/SIL/boards/1?selectedIssue=SIL-65

Changes: 
- Implemented DetailsScreenUI
- Added BottomBorderModifier util function for creating a bottom border

UI Preview: 

Screenshot 2023-02-23 at 13 50 55

Vs Figma: 

image

As you see I've changed Ui a little: 
- an increase in item height
- a different colour for the not reviewed title ( on Figma there is colour from dark mode used, so I've picked the colour from light mode). 

Let me know if you don't agree with that changes. 📣


app/src/main/java/com/appunite/loudius/domain/model/Reviewer.kt
```diff
@@ -0,0 +1,8 @@
+package com.appunite.loudius.domain.model
+
+data class Reviewer(
+    val name: String,
+    val isReviewDone: Boolean,
+    val hoursFromPRStart: Int,
+    val hoursFromReviewDone: Int?,
+)
```
Krzysiudan, line 8: **Info:** That's only the POC example model of the Reviewer which would be changed.

nowakweronika (approved):
lgtm

app/src/main/java/com/appunite/loudius/ui/theme/Color.kt
```diff
@@ -13,3 +13,6 @@ val Pink40 = Color(0xFF7D5260)
 val White99 = Color(0xFFFFFBFE)
 val Black90 = Color(0xFF1C1B1F)
 val PurpleBlack30 = Color(0xFF49454F)
+
+val NeutralVariant30 = Color(0xFF49454F)
```
kezc, line 17: thought: we need to decide (on standup :D) if we want to name colors by color name or by function.

app/src/main/java/com/appunite/loudius/ui/utils/BottomBorderModifier.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.composed
+import androidx.compose.ui.draw.drawBehind
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.unit.Dp
+
+fun Modifier.bottomBorder(strokeWidth: Dp, color: Color) = composed(
```
kezc, line 11: suggestion: have you thought about using `Divider`?

app/src/main/java/com/appunite/loudius/ui/theme/Color.kt
Krzysiudan: I totally agree with that. Firstly I wanted to stick to the convention and use a colour name, but the problem was that we have 2 different greys - neutral and neutral variant. How to name it then?

app/src/main/java/com/appunite/loudius/ui/utils/BottomBorderModifier.kt
Krzysiudan: Yes, I thought and honestly I don't have a strong opinion on which way is better.

app/src/main/java/com/appunite/loudius/ui/theme/Color.kt
Krzysiudan: We agreed to use figma names on Slack ✅

PR #11, opened 2023-02-26
Basic pull requests list

kezc:
This is very basic implementation of the PRs screen.
Screenshot 2023-02-26 at 15 24 50
It's missing 
- tests
- we always display PRs of a hard-coded user as we don't have a proper token implementation yet
- correct ui
- error handling, loading, etc.
but it displays some PRs :)

PR #15, opened 2023-02-28
[SIL-62/SIL-75] Provide new annotation for API instances

kezc:
Here we raised our concerns about how 2 retrofit instances inside hilt are implemented
https://github.com/appunite/Loudius/pull/11#discussion_r1118096637

Changes:
- 2 instead of 1 qualifier annotation, which makes programmers explicitly decide which retrofit they want to use
- some constants were renamed

What wasn't done and I don't want to change it (unless you make me :D) - splitting the network module into two modules, where one is auth module. I feel that the current solution with both `@Provides` in one file is a little better because it's easier to see that there are 2 instances of retrofit. Also, I started having a problem with the placement of GSON provider, whether should I create yet another module as it's related to both auth and not-auth networking or place it in one of them? So I decided to stick with one file. :D 

Everything should work as it was, no changes for the user


app/src/main/java/com/appunite/loudius/common/Constants.kt
```diff
@@ -2,7 +2,7 @@ package com.appunite.loudius.common
 
 object Constants {
 
-    const val BASE_URL = "https://github.com"
+    const val BASE_AUTH_URL = "https://github.com"
```
nowakweronika: suggestion: I would name it like annotations, e.g. `BASE_API_URL` and `AUTH_API_URL`. I think it will be the cleanest way.

nowakweronika (approved):

app/src/main/java/com/appunite/loudius/di/APIScope.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import javax.inject.Qualifier
+
+@Qualifier
+@Retention(AnnotationRetention.RUNTIME)
+annotation class BaseAPI
+
+@Qualifier
+@Retention(AnnotationRetention.RUNTIME)
+annotation class AuthAPI
```
Krzysiudan: **Suggestion: ** Maybe the `APIQualifiers.kt` file name would be more accurate

Krzysiudan (approved):
**Praise**: That's a lot better than before 👌

app/src/main/java/com/appunite/loudius/di/APIScope.kt
kezc: Nice catch, I was trying something different before and forgot to change 🤦

PR #17, opened 2023-03-01
[SIL-73] Storing access token

Krzysiudan:
https://silentus-au.atlassian.net/jira/software/projects/SIL/boards/1?selectedIssue=SIL-73

Changes:
- implemented storing the access token
- added junit 5 and tests for the happy path
- renamed a few components

I've decided not to implement storing access token with jetpack security because of our estimation. That could be time-consuming. Anyway if we will have free time in the end then we can try it. I've added the task to the backlog: 

https://silentus-au.atlassian.net/jira/software/projects/SIL/boards/1/backlog?selectedIssue=SIL-74


app/src/main/java/com/appunite/loudius/domain/UserRepositoryImpl.kt
```diff
@@ -1,13 +1,27 @@
...
+    override suspend fun getAndSaveAccessToken(
+        clientId: String,
+        clientSecret: String,
+        code: String,
+    ): Result<AccessToken> {
+        val tokenFromLocal = userLocalDataSource.getAccessToken()
+
+        return if (tokenFromLocal != null) {
+            // TODO: Propose removal of AccessToken data class
+            Result.success(AccessToken(tokenFromLocal))
```
Krzysiudan: **Question:** Do we need AccessToken data class which wraps only one String to our auth work?

app/src/main/java/com/appunite/loudius/domain/UserLocalDataSource.kt
```diff
@@ -0,0 +1,25 @@
...
+
+@Singleton
+class UserLocalDataSource @Inject constructor(@ApplicationContext context: Context) {
+
+    companion object {
+        private const val FILE_NAME = "com.appunite.loudius.sharedPreferences"
+        private const val KEY_ACCESS_TOKEN = "access_token"
+    }
+
+    private val sharedPreferences: SharedPreferences =
```
kezc, line 17: thought: I see in the description that you would like to try jetpack security, I think also it could be interesting to use `Preferences DataStore`, which is shared preferences successor.

app/src/main/java/com/appunite/loudius/domain/UserRepositoryImpl.kt
kezc: I don't think we need a data class for that, but have you thought maybe about using typalias? It could be interesting, but really preference based :P

app/src/test/java/com/appunite/loudius/domain/UserRepositoryImplTest.kt
```diff
@@ -0,0 +1,59 @@
...
+import io.mockk.coVerify
+import io.mockk.every
+import io.mockk.mockk
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class UserRepositoryImplTest {
```
kezc, line 14: Praise: nice tests

app/src/test/java/com/appunite/loudius/domain/UserRepositoryImplTest.kt
```diff
@@ -0,0 +1,59 @@
...
+                result,
+            ) { "Expected success result with valid access token" }
+        }
+
+    @Test
+    fun `GIVEN not saved token locally WHEN getting token THEN new token should be saved`() =
+        runTest {
+            repository.getAndSaveAccessToken("clientId", "clientSecret", "code")
+
+            coVerify(exactly = 1) { localDataSource.saveAccessToken(any()) }
```
kezc: suggestion: maybe we could check if `validAccessToken` is saved instead of any

app/src/main/java/com/appunite/loudius/domain/UserRepositoryImpl.kt
```diff
@@ -1,13 +1,27 @@
...
 
 class UserRepositoryImpl @Inject constructor(
-    private val githubDataSource: GithubDataSource,
+    private val userDataSource: UserDataSource,
+    private val userLocalDataSource: UserLocalDataSource,
 ) : UserRepository {
 
-    override suspend fun getAccessToken(clientId: String, clientSecret: String, code: String): Result<AccessToken> =
-        githubDataSource.getAccessToken(clientId, clientSecret, code)
+    override suspend fun getAndSaveAccessToken(
```
kezc: issue: what if i want to get the token in e.g. pul request details screen? Should I pass the code from login everywhere?
because this function requires as a parameter the code from the log in, it seems to be only useful for saving the token. Why would I request the code from github and not generate a new access token?
I feel like we should have one function for generating the token from code and second one for accessing the token

app/src/main/java/com/appunite/loudius/domain/UserRepositoryImpl.kt
nowakweronika: I don't think too we need a data class. First, I thought about refresh token, when I implemented it.

app/src/test/java/com/appunite/loudius/domain/UserRepositoryImplTest.kt
nowakweronika: I don't think we need to check `validAccessToken`, because we check the call count with exactly

nowakweronika (commented):
There are conflicts with develop branch.

app/src/main/java/com/appunite/loudius/domain/UserLocalDataSource.kt
Krzysiudan: We agreed on Slack - not to use it in this situation.

app/src/main/java/com/appunite/loudius/domain/UserRepositoryImpl.kt
Krzysiudan: I thought we don't need that data class but actually, we need it. Otherwise, we will get an exception with JSON parsing. It's about the structure we get from the API. 
I renamed it to `AccessTokenResponse` and then unpack it and use it as String with typealias `AccessToken`.

app/src/test/java/com/appunite/loudius/domain/UserRepositoryImplTest.kt
Krzysiudan: Good idea, after all, why not check if the value is correct too.  I've added that.

app/src/main/java/com/appunite/loudius/domain/UserRepositoryImpl.kt
Krzysiudan: Thanks for spotting that! My brain must have got some malfunctions xD 
I've added an additional function named `getAccessToken`.

app/src/main/java/com/appunite/loudius/domain/UserLocalDataSource.kt
```diff
@@ -21,5 +21,5 @@ class UserLocalDataSource @Inject constructor(@ApplicationContext context: Conte
         sharedPreferences.edit().putString(KEY_ACCESS_TOKEN, accessToken).apply()
     }
 
-    fun getAccessToken(): String? = sharedPreferences.getString(KEY_ACCESS_TOKEN, null)
+    fun getAccessToken(): String = sharedPreferences.getString(KEY_ACCESS_TOKEN, null) ?: ""
```
kezc, line 24: nitpick 1: we can use here `.orEmpty()`
nitpick 2: I don't what is your preference but I like to do something like:
```kotlin
var accessToken: String
        get() = sharedPreferences.getString(KEY_ACCESS_TOKEN, null).orEmpty()
        set(value) = sharedPreferences.edit().putString(KEY_ACCESS_TOKEN, value).apply()
```

kezc (approved):

nowakweronika (approved):

app/src/main/java/com/appunite/loudius/domain/UserLocalDataSource.kt
Krzysiudan: Looks good to me too but it would require repairing tests, and using mockks for setters and getters. It's a lot of additional work to do, so I believe it's better to leave it as it is.

PR #18, opened 2023-03-01
[SIL-66] Add client secret to build config

nowakweronika:
### Description
Added client secret key to build config variables and to github actions.
To have this variable in your system, you have to set environment variable on your mac.
To do this using zsh:
1. `Launch zsh (command zsh)`
2. `$ echo 'export CLIENT_SECRET=you know what:))' >> ~/.zshenv`
3. `$ echo $CLIENT_SECRET`
4. and restart your computer :)

### Why?
We want to keep client secret key safe.

### Documentation
[SIL-66](https://silentus-au.atlassian.net/browse/SIL-66)


app/build.gradle
```diff
@@ -21,6 +21,7 @@ android {
         vectorDrawables {
             useSupportLibrary true
         }
```
kezc, line 23: suggestion: maybe we could add the instruction to readme file so it won't get lost. Also, I think there are two mistakes:
- Launch zsh in 2nd point - I think we have to launch zsh terminal before you even run the first command :D
- 3rd point - `echo $CLIENT_SECRET`

kezc (approved):
LGTM

Krzysiudan (approved):
lgtm

app/build.gradle
jacek-marchwicki: suggestion: 
```
if (System.env.CLIENT_SECRET.isEmpty()) {
  throw RuntimeException("You need to set CLIENT_SECRET in your environment variables")
}

nowakweronika: readme updated

PR #20, opened 2023-03-02
[SIL-62] Adjust Pull Requests screen UI

kezc:
- Updated designs according to [Figma](https://www.figma.com/file/yNbcX8VTwufnf7HacYz5Zg/Material-3-Design-Kit-(Community)?node-id=53295%3A27471&t=0CyHlVhlQrk6BX9U-0)
- removed the back icon
- small change to retrieving repository name
- updated colors (there was dark mode color in light mode)

Screenshot 2023-03-02 at 10 37 18


app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -74,32 +113,32 @@ fun PullRequestsScreenPreview() {
...
+                title = "[SIL-67] Details screen - network layer",
                 updatedAt = "2021-11-29T16:31:41Z",
             ),
             PullRequest(
                 id = 1,
                 draft = true,
                 number = 1,
-                repositoryUrl = "${Constants.BASE_API_URL}/appunite/Silentus",
-                title = "PR 2",
+                repositoryUrl = "${Constants.BASE_API_URL}/repos/appunite/Silentus",
```
Krzysiudan: **Issue:** MaterialTheme closure is missing from the preview. 
That's why the preview is not showing the correct colours.

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -33,30 +41,61 @@ private fun PullRequestsScreenStateless(
...
+            .clickable(onClick = onClick),
+    ) {
+        Image(
+            painter = painterResource(id = R.drawable.ic_share),
+            contentDescription = null,
+            modifier = Modifier
+                .padding(start = 19.dp, top = 10.dp)
+                .size(width = 18.dp, height = 20.dp),
+        )
+        Column(Modifier.padding(start = 19.dp, top = 8.dp, bottom = 8.dp)) {
```
Krzysiudan: **Issue: ** It's better not to use odd numbers like 19. dp if we don't need it. 

Chat Gpt says (or you can google it): 

> Rounding odd dimension numbers to even in an Android project is a common practice because it can help prevent visual inconsistencies and improve performance on certain devices.
> 
> Firstly, many UI elements in Android use dimensions in pixels, and many device displays have a higher pixel density in one direction than the other (i.e. more pixels horizontally than vertically or vice versa). This means that using an odd number for a dimension in the direction with the higher pixel density can result in sub-pixel rendering, which can make the element look blurry or distorted. By rounding up to the nearest even number, you can ensure that the element is rendered on full pixels, resulting in a sharper and more consistent appearance.
> 
> Secondly, some Android devices use hardware acceleration to speed up the rendering of UI elements. However, some operations, such as scaling and rotation, can only be performed efficiently on even-sized images or dimensions. By using even-sized dimensions, you can help ensure that these operations are performed efficiently, leading to better performance and smoother animations.
> 
> Overall, while it may seem like a small detail, rounding odd dimension numbers to even can help improve the overall visual consistency and performance of an Android project.


So sometimes it's better to correct what is on figma.

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -33,30 +41,61 @@ private fun PullRequestsScreenStateless(
...
+            Text(text = pullRequestTitle, style = MaterialTheme.typography.bodyLarge)
+            Text(
+                text = repositoryName,
+                style = MaterialTheme.typography.bodyMedium,
+                color = MaterialTheme.colorScheme.onSurfaceVariant,
+            )
+        }
```
nowakweronika: Suggestion: I think we can extract this column to a separate method, it will cause cleaner code :)

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -33,30 +41,61 @@ private fun PullRequestsScreenStateless(
...
+        modifier = Modifier
+            .fillMaxWidth()
+            .background(backgroundColor)
+            .clickable(onClick = onClick),
+    ) {
+        Image(
+            painter = painterResource(id = R.drawable.ic_share),
+            contentDescription = null,
+            modifier = Modifier
+                .padding(start = 19.dp, top = 10.dp)
```
Krzysiudan: **Issue: **It's better not to use odd numbers like 19. dp if we don't need it.

Krzysiudan (approved):
lgtm

nowakweronika (approved):
lgtm

PR #22, opened 2023-03-06
[SIL-82] FIX - null pointer exception on getting access token. 

Krzysiudan:
https://silentus-au.atlassian.net/jira/software/projects/SIL/boards/1?selectedIssue=SIL-82

Changes: 
- wrapped getting access token from composable in LaunchedEffect
- removing code used for getting token from the intent

https://developer.android.com/jetpack/compose/side-effects - info about side effects. 

Source of the bug: 
- We shouldn't run such actions from composable without Launched effects. `Code` for getting access token is only for single use. In the previous version composable could run `getAccessToken` multiple times and then we were getting null from the network because stored `code` was used already. 

Note:
- I wasn't concerned about what should be shown on that screen cause it will be changed anyway.

### DIFF
--- a/.github/workflows/run-code-quality-check.yml
+++ b/.github/workflows/run-code-quality-check.yml
@@ -17,7 +17,11 @@ concurrency:
   group: ${{ github.ref }}-${{ github.workflow }}
   cancel-in-progress: true
 
-permissions: write-all
+permissions:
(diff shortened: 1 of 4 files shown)


nowakweronika (approved):
Nice 👌🏻

PR #24, opened 2023-03-07
[SIL-62] Fetch current user + send auth token to every request

kezc:
- Added sending auth token to every request (AuthInterceptor)
- Added okhttp logger, so we can easily see what requests are sent
- Added call to fetch current user login, so now we display logged in user's pull requests

No changes in UI : )

To test:
- [x] Log in
- [x] Check if you see your pull requests

### DIFF
--- a/app/src/main/java/com/appunite/loudius/common/ResultExtension.kt
+++ b/app/src/main/java/com/appunite/loudius/common/ResultExtension.kt
@@ -1,0 +1,4 @@
+package com.appunite.loudius.common
+
+inline fun <T, G> Result<T>.flatMap(mapper: (value: T) -> Result<G>): Result<G> =
+    fold(onSuccess = mapper, onFailure = { Result.failure(it) })
--- a/app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
+++ b/app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
@@ -2,6 +2,7 @@ package com.appunite.loudius.di
 
 import com.appunite.loudius.common.Constants
 import com.appunite.loudius.network.utils.LocalDateTimeDeserializer
+import com.appunite.loudius.network.utils.AuthInterceptor
 import com.google.gson.FieldNamingPolicy
 import com.google.gson.Gson
 import com.google.gson.GsonBuilder
@@ -21,12 +22,8 @@ import retrofit2.converter.gson.GsonConverterFactory
 object NetworkModule {
     @Provides
     @Singleton
-    fun provideOkHttp(): OkHttpClient {
-        val logging = HttpLoggingInterceptor()
-        logging.setLevel(HttpLoggingInterceptor.Level.BASIC)
-        return OkHttpClient.Builder()
-            .addInterceptor(logging)
-            .build()
+    fun provideLoggingInterceptor(): HttpLoggingInterceptor = HttpLoggingInterceptor().apply {
+        level = HttpLoggingInterceptor.Level.BASIC
     }
 
     @Provides
@@ -43,13 +40,18 @@ object NetworkModule {
     fun provideAuthRetrofit(
         gson: Gson,
         @AuthAPI baseUrl: String,
-        okHttpClient: OkHttpClient
-    ): Retrofit =
-        Retrofit.Builder()
+        loggingInterceptor: HttpLoggingInterceptor
+    ): Retrofit {
+        val okHttpClient = OkHttpClient.Builder()
+            .addInterceptor(loggingInterceptor)
+            .build()
+
+        return Retrofit.Builder()
             .baseUrl(baseUrl)
             .client(okHttpClient)
             .addConverterFactory(GsonConverterFactory.create(gson))
             .build()
+    }
 
     @Provides
     @Singleton
@@ -57,13 +59,20 @@ object NetworkModule {
     fun provideBaseRetrofit(
         gson: Gson,
         @BaseAPI baseAPIUrl: String,
-        okHttpClient: OkHttpClient
-    ): Retrofit =
-        Retrofit.Builder()
+        loggingInterceptor: HttpLoggingInterceptor,
+        authInterceptor: AuthInterceptor
+    ): Retrofit {
+        val okHttpClient = OkHttpClient.Builder()
+            .addInterceptor(authInterceptor)
+            .addInterceptor(loggingInterceptor)
+            .build()
+
+        return Retrofit.Builder()
             .baseUrl(baseAPIUrl)
             .client(okHttpClient)
             .addConverterFactory(GsonConverterFactory.create(gson))
             .build()
+    }
 
     @Provides
     @Singleton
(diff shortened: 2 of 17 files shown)


app/src/main/java/com/appunite/loudius/ui/repos/ReposScreen.kt
```diff
@@ -22,6 +24,8 @@ fun ReposScreen(
         rememberedCode.value?.let {
             viewModel.getAccessToken(it)
             intent.data = null
+            delay(1000)
```
kezc, line 27: temporary solution until we have a proper loading screen.....

app/src/main/java/com/appunite/loudius/network/services/GithubPullRequestsService.kt
```diff
@@ -22,14 +22,16 @@ interface GithubPullRequestsService {
...
     suspend fun getReviews(
         @Path("owner") owner: String,
         @Path("repo") repo: String,
         @Path("pull_number") pullRequestNumber: String,
-        @Header("Authorization") token: String,
     ): List<Review>
+
+    @Headers("Accept: application/json")
+    @GET("user")
+    suspend fun getUser(): User
```
kezc: This method probably should not belong here. I'll change it in a different PR, as this might require a lot of renaming and a large number of new classes
I have already created a ticket so I don't forget :P https://silentus-au.atlassian.net/jira/software/projects/SIL/boards/1?selectedIssue=SIL-86

app/src/main/java/com/appunite/loudius/common/ResultExtension.kt
```diff
@@ -0,0 +1,4 @@
+package com.appunite.loudius.common
+
+inline fun <T, G> Result<T>.flatMap(mapper: (value: T) -> Result<G>): Result<G> =
+    fold(onSuccess = mapper, onFailure = { Result.failure(it) })
```
Krzysiudan, line 4: **Praise:** Nice idea with that function 👌 
**Suggestion: ** I'm not so sure if it should be named `flatMap`, which is usually used with collections or sequences. So it can be at first sight a little bit misleading. Maybe `mapResult` or "mapValue". Change is up to you.

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
```diff
@@ -17,6 +20,11 @@ import javax.inject.Singleton
 @InstallIn(SingletonComponent::class)
 @Module
 object NetworkModule {
+    @Provides
+    @Singleton
+    fun provideLoggingInterceptor(): HttpLoggingInterceptor = HttpLoggingInterceptor().apply {
+        level = HttpLoggingInterceptor.Level.BASIC
+    }
```
Krzysiudan, line 27: **Praise:** Great idea 👌

app/src/main/java/com/appunite/loudius/network/model/User.kt
```diff
@@ -0,0 +1,3 @@
+package com.appunite.loudius.network.model
+
+data class User(val id: Int, val login: String)
```
Krzysiudan, line 3: **nitpick: ** I wonder if we want to use a user avatar. Then we could add the `avatar_url` too.

app/src/main/java/com/appunite/loudius/network/services/GithubPullRequestsService.kt
```diff
@@ -22,14 +22,16 @@ interface GithubPullRequestsService {
...
 
     @GET("/repos/{owner}/{repo}/pulls/{pull_number}/reviews")
     suspend fun getReviews(
         @Path("owner") owner: String,
         @Path("repo") repo: String,
         @Path("pull_number") pullRequestNumber: String,
-        @Header("Authorization") token: String,
     ): List<Review>
+
+    @Headers("Accept: application/json")
```
Krzysiudan: **Question:** It won't work without that header? I'm just curious.

app/src/main/java/com/appunite/loudius/network/utils/AuthInterceptor.kt
```diff
@@ -0,0 +1,17 @@
...
+class AuthInterceptor @Inject constructor(
+    private val userRepository: UserRepository,
+) : Interceptor {
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val authenticatedRequest = chain.request().newBuilder()
+            .addHeader("Authorization", "Bearer ${userRepository.getAccessToken()}")
+            .build()
+        return chain.proceed(authenticatedRequest)
+    }
+}
```
Krzysiudan, line 17: **Praise:** I like that 👌

app/src/test/java/com/appunite/loudius/fakes/FakeUserRepository.kt
```diff
@@ -0,0 +1,18 @@
...
+        clientId: String,
+        clientSecret: String,
+        code: String,
+    ): Result<AccessToken> {
+        return Result.success("validToken")
+    }
+
+    override fun getAccessToken(): String {
+        return "validToken"
+    }
```
Krzysiudan, line 17: **Suggestion: ** You can use the expression body here. Although if you want later add when expression then it's probably a waste of time.

app/src/test/java/com/appunite/loudius/network/AuthInterceptorTest.kt
```diff
@@ -0,0 +1,51 @@
...
+            val successResponse = MockResponse().setBody(testDataJson)
+            mockWebServer.enqueue(successResponse)
+
+            service.test()
+            val request = mockWebServer.takeRequest()
+            val header = request.getHeader("authorization")
+
+            assertEquals("Bearer validToken", header)
+        }
+    }
```
Krzysiudan, line 42: **Praise:** Nice conception with that Interceptor test in separation.

Krzysiudan (approved):
LGTM. The limit of praises for a month has been exceeded. xD

app/src/main/java/com/appunite/loudius/common/ResultExtension.kt
kezc: `flatMap` is a different name for monad's bind function which comes from functional programming https://en.wikipedia.org/wiki/Monad_(functional_programming)#Definition List and sequences are monads and that's why they use this name too

app/src/main/java/com/appunite/loudius/network/model/User.kt
kezc: I think we can add it to the class when it is actually necessary. I don't have any use case for that in my code, so I don't want to be responsible for it :D

app/src/main/java/com/appunite/loudius/network/services/GithubPullRequestsService.kt
kezc: it works, I'll remove it

app/src/test/java/com/appunite/loudius/fakes/FakeUserRepository.kt
kezc: In the back of my mind, I thought it might be extended in the future, so I decided to use brackets 😄

app/src/main/java/com/appunite/loudius/common/ResultExtension.kt
Krzysiudan: Then thanks for the explanation, you are right. I didn't know that 👌

PR #25, opened 2023-03-07
[SIL-84] Reviewers screen logic

Krzysiudan:
Details screen logic - view model implementation

Changes: 
- implemented `RevieversViewModel
- implemented navigation to RevieversScreen from PullRequestScreen with passing data
- changed PullRequest data class to get `createdAt` instead of `updatedAt` and store it as LocalDateTime
- changed particular user id to stubed one. 

Screenshot 2023-03-09 at 10 22 54
Right now as you can see user is not excluded from the reviewers. I've created separate task for that: 
https://silentus-au.atlassian.net/jira/software/projects/SIL/boards/1?selectedIssue=SIL-88

- [x] Check if you can see reviewers for your PR.


app/src/main/java/com/appunite/loudius/MainActivity.kt
```diff
@@ -16,6 +16,7 @@ import com.appunite.loudius.common.Screen
 import com.appunite.loudius.ui.login.LoginScreen
 import com.appunite.loudius.ui.pullrequests.PullRequestsScreen
 import com.appunite.loudius.ui.repos.ReposScreen
+import com.appunite.loudius.ui.reviewers.ReviewersScreen
```
Krzysiudan, line 19: **info** That's a comment for a test.

app/src/main/java/com/appunite/loudius/MainActivity.kt
```diff
@@ -51,7 +52,21 @@ class MainActivity : ComponentActivity() {
...
+                                    submissionDate = submissionTime,
+                                )
+                                navController.navigate(route)
+                            }
+                        }
+                        composable(
+                            route = Screen.Reviewers.route,
+                            arguments = Screen.Reviewers.arguments,
+                        ) {
+                            ReviewersScreen({ navController.popBackStack() })
```
nowakweronika: suggestion: I would make named parameter for better readability.

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -30,16 +30,24 @@ import com.appunite.loudius.common.Constants
 import com.appunite.loudius.network.model.PullRequest
 import com.appunite.loudius.ui.components.LoudiusTopAppBar
 import com.appunite.loudius.ui.theme.LoudiusTheme
+import java.time.LocalDateTime
 
 @Composable
-fun PullRequestsScreen(viewModel: PullRequestsViewModel = hiltViewModel()) {
+fun PullRequestsScreen(
+    viewModel: PullRequestsViewModel = hiltViewModel(),
+    navigateToReviewers: (String, String, String, String) -> Unit,
```
nowakweronika: suggestion: Maybe we can make `typealias NavigateToReviewers = (String, String, String, String) -> Unit` for cleaner code :)

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -30,16 +30,24 @@ import com.appunite.loudius.common.Constants
...
+    PullRequestsScreenStateless(
+        pullRequests = state.pullRequests,
+        onItemClick = navigateToReviewers,
+    )
 }
 
 @Composable
 private fun PullRequestsScreenStateless(
     pullRequests: List<PullRequest>,
+    onItemClick: (String, String, String, String) -> Unit,
```
nowakweronika: suggestion: I would name it `onPullRequestItemClick` or something like that

nowakweronika (approved):

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
```diff
@@ -21,31 +21,54 @@ import androidx.compose.ui.res.painterResource
...
+import androidx.hilt.navigation.compose.hiltViewModel
 import com.appunite.loudius.R
 import com.appunite.loudius.domain.model.Reviewer
 import com.appunite.loudius.ui.components.LoudiusTopAppBar
 import com.appunite.loudius.ui.theme.LoudiusTheme
 import com.appunite.loudius.ui.utils.bottomBorder
 
+@Composable
+fun ReviewersScreen(
+    navigateBack: () -> Unit,
```
kezc: nitpick: if you place this parameter as last then you will be able to do trailing lambda in `MainActivity`

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
```diff
@@ -0,0 +1,100 @@
...
+    private fun getInitialValues(savedStateHandle: SavedStateHandle) = InitialValues(
+        checkNotNull(savedStateHandle[Screen.Reviewers.ownerArg]),
+        checkNotNull(savedStateHandle[Screen.Reviewers.repoArg]),
+        checkNotNull(savedStateHandle[Screen.Reviewers.pullRequestNumberArg]),
+        checkNotNull(savedStateHandle[Screen.Reviewers.submissionDateArg]),
+    )
+
+    private suspend fun fetchRequestedReviewers(initialValues: InitialValues) {
+        val (owner, repo, pullRequestNumber, submissionTime) = initialValues
+        val hoursFromPRStart = countHoursTillNow(LocalDateTime.parse(submissionTime))
```
kezc: nitpick: this probably could've been placed in `mapToReviewers`

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
```diff
@@ -0,0 +1,100 @@
...
+        private set
+
+    init {
+        val initialValues = getInitialValues(savedStateHandle)
+
+        state = state.copy(pullRequestNumber = initialValues.pullRequestNumber)
+
+        viewModelScope.launch {
+            fetchRequestedReviewers(initialValues)
+            fetchReviews(initialValues)
```
kezc, line 41: thought: I would rethink the current approach to fetching data. I feel that displaying errors/loading might be a little bit complicated because you have 2 requests running in 2 functions, so you will have to probably return some value indicating if it was successful. Maybe it would be better to have 3 functions
- 2 functions that fetch, map, and return data
- one function which combines this data, orchestrates loading and error handling
- in addition, this would make running requests simultaneously easier, as they are currently called sequentially.

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -78,14 +84,28 @@ private fun PullRequestItem(
...
 
+private fun onItemClick(
+    onClick: (String, String, String, String) -> Unit,
+    data: PullRequest,
+) {
+    onClick(
+        data.owner,
+        data.shortRepositoryName,
+        data.number.toString(),
+        LocalDateTime
```
kezc: suggestion: can't we get current time in `ReviewersViewModel`?

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -52,10 +60,9 @@ private fun PullRequestsScreenStateless(
             itemsIndexed(pullRequests) { index, item ->
                 val isIndexEven = index % 2 == 0
                 PullRequestItem(
-                    repositoryName = item.fullRepositoryName,
-                    pullRequestTitle = item.title,
+                    data = item,
```
kezc, line 72: suggestion: we pass the whole item, so we can handle onClick. Maybe it would be better to pass the item id and handle it through the view model, which would emit the corresponding item.

app/src/main/java/com/appunite/loudius/network/model/PullRequest.kt
```diff
@@ -13,6 +13,12 @@ data class PullRequest(
     val fullRepositoryName: String
```
kezc, line 14: issue: I can't write a comment on line 11,  so I place it here, but I refer to this specific line.
In the description, you say
`changed PullRequest data class to get createdAt instead of updatedAt and store it as LocalDateTime`
There is no such change

Krzysiudan: Heh, I've added that change at the end and forgotten to push it. 😞

app/src/main/java/com/appunite/loudius/MainActivity.kt
Krzysiudan: I've reverted the parameter order in the `ReviewersScreen` constructor and now lambda is outside of the parenthesis and It's impossible to name it now.

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
Krzysiudan: Good idea, I've done that👌

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
Krzysiudan: Hmm, when there is no other on-click and no other item types, I think the shorter the better.

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
Krzysiudan: 1. Do you mean launching these requests in separate coroutines? 🤔 

You are probably right in that statement that it could be a bit complicated to implement here loading and error handling. 
I propose that I will refactor it in a way it's easier with error handling and loading in time when I will implement these functionalities. Otherwise, I could end with refactoring this once more when obstacles appear. Is it okay for you?

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
Krzysiudan: Heh, I've added that change at the end and forgotten to push it. 😞

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
Krzysiudan: Do you feel that passing this much of the data to the item is bad for performance or it's bad in a different way? 🤔 

Maybe it's a good topic to think about it wider because I'm starting to wonder too. Should we make such navigation events through the view model or not? Right now view model is omitted.

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
kezc: sure, I just wanted to point out a possible problem 😄

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
Krzysiudan: In favour of moving that to the view model is definitely that it is possible to test navigation by a unit test.

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
```diff
@@ -51,17 +51,20 @@ class ReviewersViewModel @Inject constructor(
...
-                val reviewers = response.mapToReviewers(hoursFromPRStart)
+                val reviewers = response.mapToReviewers(submissionTime)
                 state = state.copy(reviewers = state.reviewers + reviewers)
             }
     }
 
-    private fun RequestedReviewersResponse.mapToReviewers(hoursFromPRStart: Long) = users.map {
-        Reviewer(it.id, it.login, false, hoursFromPRStart, null)
+    private fun RequestedReviewersResponse.mapToReviewers(submissionTime: String): List<Reviewer> {
+        val hoursFromPRStart = countHoursTillNow(LocalDateTime.parse(submissionTime))
```
kezc, line 63: suggestion: I thought that you would do it in both mappers :D

Krzysiudan: What can I say? You were in a big mistake. xD

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
Krzysiudan: I've moved the responsibility of the navigation to the view model.

kezc (approved):
lgtm

PR #27, opened 2023-03-08
[SIL-86] Rename user and auth network classes

kezc:
This is my proposal for splitting network parts, however, I'm not so sure if it's a good approach.
The problem raises from the fact, that auth requests require a different service than the rest of the calls, as it has a different API URL. Due to that, I couldn't place `getUser` inside `UserService`, so I decided to rename it to `AuthService` and create a new `GithubUserService`. As a chain of events, this resulted in renaming data sources and repositories too.

In conclusion:
GithubApi -> GithubAuthService
UserRepository -> AuthRepository
UserDataSource -> AuthDataSource 

new -> UserDataSource
new -> GithubUserService

### DIFF
--- a/app/src/main/java/com/appunite/loudius/di/GithubModule.kt
+++ b/app/src/main/java/com/appunite/loudius/di/GithubModule.kt
@@ -1,9 +1,9 @@
 package com.appunite.loudius.di
 
 import android.content.Context
-import com.appunite.loudius.domain.UserLocalDataSource
 import com.appunite.loudius.domain.AuthRepository
 import com.appunite.loudius.domain.AuthRepositoryImpl
+import com.appunite.loudius.domain.UserLocalDataSource
 import com.appunite.loudius.network.datasource.AuthDataSource
 import com.appunite.loudius.network.datasource.AuthNetworkDataSource
 import com.appunite.loudius.network.services.AuthService
--- a/app/src/main/java/com/appunite/loudius/di/GithubModule.kt
+++ b/app/src/main/java/com/appunite/loudius/di/GithubModule.kt
@@ -2,12 +2,13 @@ package com.appunite.loudius.di
 
 import android.content.Context
 import com.appunite.loudius.domain.UserLocalDataSource
-import com.appunite.loudius.domain.UserRepository
-import com.appunite.loudius.domain.UserRepositoryImpl
-import com.appunite.loudius.network.UserDataSource
-import com.appunite.loudius.network.UserNetworkDataSource
-import com.appunite.loudius.network.services.GithubApi
+import com.appunite.loudius.domain.AuthRepository
+import com.appunite.loudius.domain.AuthRepositoryImpl
+import com.appunite.loudius.network.datasource.AuthDataSource
+import com.appunite.loudius.network.datasource.AuthNetworkDataSource
+import com.appunite.loudius.network.services.GithubAuthService
 import com.appunite.loudius.network.services.GithubPullRequestsService
+import com.appunite.loudius.network.services.GithubUserService
 import dagger.Module
 import dagger.Provides
 import dagger.hilt.InstallIn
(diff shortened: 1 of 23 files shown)


app/src/main/java/com/appunite/loudius/di/GithubModule.kt
```diff
@@ -22,25 +23,31 @@ object GithubModule {
 
     @Singleton
     @Provides
-    fun provideGithubApi(@AuthAPI retrofit: Retrofit): GithubApi =
-        retrofit.create(GithubApi::class.java)
+    fun provideGithubAuthService(@AuthAPI retrofit: Retrofit): GithubAuthService =
+        retrofit.create(GithubAuthService::class.java)
```
Krzysiudan: **Suggestion:** Removing the prefix `Github` from everywhere. 
I am not sure now if this prefix gives us additional value. There is no other API then GitHub API.  
What do you think? Do you consider it helpful? 

I mean to rename components: 
`GithubAuthService` -> `AuthService`
`GithubUserService` -> `UserService`
`GithubPullRequestsService` -> `PullRequestsService' 
etc..

app/src/main/java/com/appunite/loudius/network/datasource/UserDataSource.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import com.appunite.loudius.network.model.User
+import com.appunite.loudius.network.services.GithubUserService
+import com.appunite.loudius.network.utils.safeApiCall
+import javax.inject.Inject
+
+class UserDataSource @Inject constructor(private val userService: GithubUserService) {
+    suspend fun getUser(): Result<User> = safeApiCall {
+        userService.getUser()
+    }
```
Krzysiudan: **Question: ** Tests in the separate PR right? 😅

app/src/main/java/com/appunite/loudius/di/GithubModule.kt
```diff
@@ -22,25 +23,31 @@ object GithubModule {
...
+    ): AuthRepository = AuthRepositoryImpl(authDataSource, userLocalDataSource)
 
     @Singleton
     @Provides
-    fun provideUserDataSource(
-        api: GithubApi,
-    ): UserDataSource = UserNetworkDataSource(api)
+    fun provideAuthServiceDataSource(
+        service: GithubAuthService,
+    ): AuthDataSource = AuthNetworkDataSource(service)
```
Krzysiudan: **Steve Wonder:** I'm wondering if we want in same module dependencies for services and repos/data sources.
What opinions do you have? 🤔

Krzysiudan (approved):
Lgtm. That's cleaner for sure. 👌

app/src/main/java/com/appunite/loudius/di/GithubModule.kt
nowakweronika: Both forms are ok to me :)

app/src/main/java/com/appunite/loudius/di/GithubModule.kt
nowakweronika: It's ok to me, I don't like too much modules in small project, I don't see benefits from it.

app/src/main/java/com/appunite/loudius/network/datasource/UserDataSource.kt
kezc: of course 😆

app/src/main/java/com/appunite/loudius/di/GithubModule.kt
kezc: maybe we could separate it by layer, but I don't have a strong opinion on that. Maybe I could add a ticket so we can think about it before showing publicly our app or if someone will have nothing to do?

app/src/main/java/com/appunite/loudius/di/GithubModule.kt
kezc: I fully agree, it annoys me :P I have added these changes to this PR, however, I have left GithubModule as it is.

nowakweronika (approved):
lgtm

app/src/main/java/com/appunite/loudius/di/GithubModule.kt
Krzysiudan: Nice 👌

app/src/main/java/com/appunite/loudius/di/GithubModule.kt
Krzysiudan: Creating a ticket for that sounds reasonable 👌

PR #28, opened 2023-03-09
[SIL-80] Login screen error handling

nowakweronika:
### Description
*Added error handling to login screen.*

### Why?
*We want to show error screen when something goes wrong.*

### Documentation
[**SIL-80**](https://silentus-au.atlassian.net/browse/SIL-80)

## Checklist

- [ ] To simulate error, you can i.e. add some text in `LoadingViewModel` in line 41 to `code` (i.e. `getAccessToken(it + "1")`) value and check if error screen is showing in the good way and after click is loading again and go to Repositories screen.


app/src/main/java/com/appunite/loudius/ui/loading/LoadingScreen.kt
```diff
@@ -0,0 +1,65 @@
...
+    intent: Intent,
+    viewModel: LoadingViewModel = hiltViewModel(),
+    onNavigateToPullRequest: () -> Unit,
+) {
+    viewModel.state.let { state ->
+        val code = intent.data?.getQueryParameter("code")
+        val rememberedCode = rememberUpdatedState(newValue = code)
+        LaunchedEffect(key1 = rememberedCode) {
+            rememberedCode.value?.let {
+                viewModel.getAccessToken(it)
```
kezc: suggestion: I think we should move it to view model's init block

app/src/main/java/com/appunite/loudius/ui/loading/LoadingScreen.kt
```diff
@@ -0,0 +1,65 @@
...
+import kotlinx.coroutines.delay
+
+@Composable
+fun LoadingScreen(
+    intent: Intent,
+    viewModel: LoadingViewModel = hiltViewModel(),
+    onNavigateToPullRequest: () -> Unit,
+) {
+    viewModel.state.let { state ->
+        val code = intent.data?.getQueryParameter("code")
```
kezc: suggestion: maybe we could just use `savedStateHandle` inside view model to get this variable, so we don't have to pass it everywhere

app/src/main/java/com/appunite/loudius/network/model/AccessTokenResponse.kt
```diff
@@ -1,8 +1,7 @@
 package com.appunite.loudius.network.model
 
-typealias AccessToken = String
-
 data class AccessTokenResponse(
 
-    val accessToken: String,
+    val accessToken: String?,
+    val error: String? = null,
```
kezc: nitpick: line 4: unnecessary line break
question: the error field is unused. Are you sure that it should be here?

app/src/main/java/com/appunite/loudius/ui/loading/LoadingScreen.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.appunite.loudius.R
+import com.appunite.loudius.ui.components.LoudiusErrorScreen
+import kotlinx.coroutines.delay
+
+@Composable
+fun LoadingScreen(
+    intent: Intent,
+    viewModel: LoadingViewModel = hiltViewModel(),
+    onNavigateToPullRequest: () -> Unit,
+) {
```
Krzysiudan, line 19: **Chore:** Screen preview would be appreciated.

app/src/main/java/com/appunite/loudius/ui/loading/LoadingScreen.kt
Krzysiudan: What do you mean by that? Do you mean to set the value here in savedStateHandle and get it from the view model? It could also be done by some function on the view model, without using the savedStateHandle.

app/src/main/java/com/appunite/loudius/ui/loading/LoadingScreen.kt
Krzysiudan: We would need to inject the code somehow into the view model constructor to make this in the init block. Wouldn't it be too much complicated? Or how can we make it another way?

app/src/main/java/com/appunite/loudius/network/model/AccessTokenResponse.kt
nowakweronika: We decided to leave this error field.

app/src/main/java/com/appunite/loudius/ui/loading/LoadingScreen.kt
nowakweronika: I think screen preview can be implement when loading indicator will be implemented.

Krzysiudan: Ahh okay, because now there is nothing much to show.

app/src/main/java/com/appunite/loudius/ui/loading/LoadingScreen.kt
```diff
@@ -0,0 +1,63 @@
...
+import com.appunite.loudius.R
+import com.appunite.loudius.ui.components.LoudiusErrorScreen
+
+@Composable
+fun LoadingScreen(
+    intent: Intent,
+    viewModel: LoadingViewModel = hiltViewModel(),
+    onNavigateToPullRequest: () -> Unit,
+) {
+    viewModel.state.let { state ->
```
Krzysiudan: **suggestion:** It doesn't have to be wrapped in `viewModel.state.let {` 
I believe it's more readable when there are no additional brackets. You can make a `state` variable.

app/src/main/java/com/appunite/loudius/ui/loading/LoadingViewModel.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    var state by mutableStateOf(LoadingState())
+        private set
+
+    fun setCodeAndGetAccessToken(code: String?) {
+        state = state.copy(code = code)
+        code?.let {
+            getAccessToken(it)
+        }
+    }
```
Krzysiudan: **Chore** This function should get the code as not nullable String. 
Then it would be simplified.

app/src/main/java/com/appunite/loudius/ui/loading/LoadingViewModel.kt
```diff
@@ -0,0 +1,81 @@
...
+                code = code,
+            ).onSuccess { token ->
+                state = if (token.accessToken != null) {
+                    state.copy(
+                        accessToken = token.accessToken,
+                        navigateToPullRequests = NavigateToPullRequests,
+                    )
+                } else {
+                    state.copy(showErrorScreen = true)
+                }
```
Krzysiudan: **Suggestion** Maybe it would be beneficial to extract that to a separate function like `handleGetAccessTokenSuccess`.

app/src/test/java/com/appunite/loudius/ui/loading/LoadingViewModelTest.kt
```diff
@@ -0,0 +1,74 @@
...
+        // given
+        val action = LoadingAction.OnTryAgainClick
+        viewModel.setCodeAndGetAccessToken(EXAMPLE_CODE)
+
+        // when
+        viewModel.onAction(action)
+
+        // then
+        assert(!viewModel.state.showErrorScreen)
+        assert(viewModel.state.accessToken == EXAMPLE_ACCESS_TOKEN)
```
Krzysiudan: **Question:** Do you think it's beneficial to add such comments? What does it give us?

app/src/test/java/com/appunite/loudius/ui/loading/LoadingViewModelTest.kt
```diff
@@ -0,0 +1,74 @@
...
+        private const val EXAMPLE_CODE = "code"
+        private const val EXAMPLE_ACCESS_TOKEN = "validToken"
+    }
+
+    private val repository: FakeAuthRepository = FakeAuthRepository()
+    private lateinit var viewModel: LoadingViewModel
+
+    @BeforeEach
+    fun setup() {
+        mockkStatic(Clock::class)
```
Krzysiudan: **chore:** `mockkStatic` is not needed here. 
It's only useful in places where we use LocalDateTime.now for example.

app/src/test/java/com/appunite/loudius/ui/loading/LoadingViewModelTest.kt
```diff
@@ -0,0 +1,74 @@
...
+        viewModel = LoadingViewModel(repository)
+    }
+
+    @Test
+    fun `GIVEN example valid code WHEN setCodeAndGetAccessToken THEN set code, access token and navigateToPullRequests`() {
```
Krzysiudan: **Suggestion: ** Remove word example
I believe the word `example` not gives us too much. A valid code is a valid code.

app/src/test/java/com/appunite/loudius/ui/loading/LoadingViewModelTest.kt
```diff
@@ -0,0 +1,74 @@
...
+        // when
+        viewModel.setCodeAndGetAccessToken(EXAMPLE_CODE)
+
+        // then
+        assert(viewModel.state.code == EXAMPLE_CODE)
+        assert(viewModel.state.accessToken == EXAMPLE_ACCESS_TOKEN)
+        assert(viewModel.state.navigateToPullRequests == NavigateToPullRequests)
```
Krzysiudan: **Chore:** Use `assertEquals` instead of `assert`. 

When we use assertEquals we get more info when test fails: 
Screenshot 2023-03-15 at 09 22 26

versus assert fail: 
Screenshot 2023-03-15 at 09 23 00

app/src/test/java/com/appunite/loudius/ui/loading/LoadingViewModelTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Test
+    fun `GIVEN null as code WHEN setCodeAndGetAccessToken THEN set null as code`() {
+        // when
+        viewModel.setCodeAndGetAccessToken(null)
+
+        // then
+        assert(viewModel.state.code == null)
+    }
```
Krzysiudan: **Suggestion** If you make the setCodeAndGetAccessToken method take not a nullable string then this test can be removed.

app/src/test/java/com/appunite/loudius/ui/loading/LoadingViewModelTest.kt
nowakweronika: For me it's more readable code, but I don't have strong opinion about it

app/src/test/java/com/appunite/loudius/ui/loading/LoadingViewModelTest.kt
nowakweronika: Thanks, I forgot to remove it :)

app/src/test/java/com/appunite/loudius/ui/loading/LoadingViewModelTest.kt
Krzysiudan: The new line after each of the parts serves the same goal. But we can wait till @kezc shares his opinion too.

Krzysiudan:
You can try to squash the commits in the git history. Are you sure you want to add the task number before each of commit message? I and @kezc are not doing that so it does not look consistent.

app/src/test/java/com/appunite/loudius/ui/loading/LoadingViewModelTest.kt
Krzysiudan: And it's 3 or 2 extra lines to every test. More there are sometimes cases when it's not obvious what is given, and what is when.

Krzysiudan: I would opt for removing that 😅

kezc: I don't mind it :D

nowakweronika:
> You can try to squash the commits in the git history. Are you sure you want to add the task number before each of commit message? I and @kezc are not doing that so it does not look consistent.

It was helpful, when I had to know exactly which tasks were merged in the feature, but I can stop it :)

Krzysiudan (approved):
lgtm

PR #29, opened 2023-03-10
[SIL-86] Add fetching pull requests test

kezc:
- Added tests for `PullRequestDataSource#getPullRequestsForUser`,
- changed `PullRequest#updatedAt` type to LocalDateTime so it matches the rest of our code

### DIFF
--- a/app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
+++ b/app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
@@ -31,7 +31,6 @@ import com.appunite.loudius.network.model.PullRequest
 import com.appunite.loudius.ui.components.LoudiusTopAppBar
 import com.appunite.loudius.ui.theme.LoudiusTheme
 import java.time.LocalDateTime
-import java.time.ZonedDateTime
 
 @Composable
 fun PullRequestsScreen(viewModel: PullRequestsViewModel = hiltViewModel()) {
--- a/app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
+++ b/app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
@@ -30,6 +30,7 @@ import com.appunite.loudius.common.Constants
 import com.appunite.loudius.network.model.PullRequest
 import com.appunite.loudius.ui.components.LoudiusTopAppBar
 import com.appunite.loudius.ui.theme.LoudiusTheme
+import java.time.LocalDateTime
 import java.time.ZonedDateTime
 
 @Composable
@@ -130,7 +131,7 @@ fun PullRequestsScreenPreview() {
                     number = 0,
                     repositoryUrl = "${Constants.BASE_API_URL}/repos/appunite/Stefan",
                     title = "[SIL-67] Details screen - network layer",
-                    updatedAt = ZonedDateTime.parse("2023-03-07T09:24:24Z").toLocalDateTime(),
+                    updatedAt = LocalDateTime.parse("2023-03-07T09:24:24"),
                 ),
                 PullRequest(
                     id = 1,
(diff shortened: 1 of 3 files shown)


app/src/test/java/com/appunite/loudius/network/datasource/PullRequestsNetworkDataSourceTest.kt
```diff
@@ -97,7 +258,8 @@ class PullRequestsNetworkDataSourceTest {
                     "exampleOwner",
                     "exampleRepo",
                     "exampleNumber",
-                )
+
```
kezc: suggestion: unnecessary line break

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -129,31 +130,31 @@ fun PullRequestsScreenPreview() {
                     number = 0,
                     repositoryUrl = "${Constants.BASE_API_URL}/repos/appunite/Stefan",
                     title = "[SIL-67] Details screen - network layer",
-                    updatedAt = "2021-11-29T16:31:41Z",
+                    updatedAt = ZonedDateTime.parse("2023-03-07T09:24:24Z").toLocalDateTime(),
```
Krzysiudan: **Suggestion :** You can make it shorter by :
`
LocalDateTime.parse("2023-03-07T09:24:24")
`
In other places too.

app/src/test/java/com/appunite/loudius/network/datasource/PullRequestsNetworkDataSourceTest.kt
```diff
@@ -18,30 +20,186 @@ import org.junit.jupiter.api.Assertions
...
+                        ),
+                    ),
+                ),
+            )
+
+            assertEquals(expected, actualResponse) { "Data should be valid" }
+        }
+
+        @Test
+        fun `Given incorrect access token WHEN processing request THEN return failure with Network error`() =
```
Krzysiudan: **Issue:** Name states it returns network error but test states it expects UnknownError. 
Please fix that in other tests in this file. I've made that mistake 😅

app/src/test/java/com/appunite/loudius/network/datasource/PullRequestsNetworkDataSourceTest.kt
```diff
@@ -225,7 +384,7 @@ class PullRequestsNetworkDataSourceTest {
                             "1",
                             User(33498031, "exampleUser"),
                             ReviewState.COMMENTED,
-                            LocalDateTime.parse("2023-03-02T10:21:36"),
+                            ZonedDateTime.parse("2023-03-02T10:21:36Z").toLocalDateTime(),
```
Krzysiudan: **Question:** Why did you change that? 
I don't think that we need to show in expected data that we are getting that from the `ZonedDateTime`. It's only about making correct data.

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
kezc: I prefer to have the same format as API or have it in a human-readable way :| I don't like that we remove the 'z' and have yet another date format in the app

app/src/test/java/com/appunite/loudius/network/datasource/PullRequestsNetworkDataSourceTest.kt
kezc: we probably should map this error to some UnauthorizedError or something instead, because this is a known error :P WDYT?

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
Krzysiudan: What do you mean by that? Do you recommend using ZonedDateTime all over the app? 
Here, after all, you have LocalDateTime anyway and it is without `Z`. So it's only a longer way of creating the same.

app/src/test/java/com/appunite/loudius/network/datasource/PullRequestsNetworkDataSourceTest.kt
Krzysiudan: Yeah, yeah definitely. But it's more like an error-handling topic, so the task is for later. But if you want to implement that now, then why not.

Krzysiudan: You forgot to correct the name here.

Krzysiudan (approved):
Approved but one name wasn't corrected 😅

PR #30, opened 2023-03-10
[SIL-86] Add user data source tests

kezc:
- Added user data source tests

### DIFF
--- a/app/src/test/java/com/appunite/loudius/network/datasource/UserDataSourceTest.kt
+++ b/app/src/test/java/com/appunite/loudius/network/datasource/UserDataSourceTest.kt
@@ -41,7 +41,6 @@ class UserDataSourceTest {
             ) { "Exception thrown should be NetworkError type" }
         }
 
-
     @Test
     fun `Given correct params WHEN successful response THEN return success`() = runTest {
         //language=JSON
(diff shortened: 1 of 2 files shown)


app/src/test/java/com/appunite/loudius/network/datasource/UserDataSourceTest.kt
```diff
@@ -0,0 +1,127 @@
...
+                id = 1,
+                login = "exampleUser",
+            )
+        )
+
+        Assertions.assertEquals(expected, actualResponse) { "Data should be valid" }
+    }
+
+    @Test
+    fun `Given incorrect access token WHEN processing request THEN return failure with Network error`() =
```
Krzysiudan: **Issue:** Name states it returns Network error but tests state it expects UnknownError.

Krzysiudan: The name is not corrected here.

Krzysiudan (approved):
Approved but one name is not corrected yet 😅

nowakweronika (commented):
Conflict with `PullRequestsNetworkDataSourceTest.kt` :)

PR #32, opened 2023-03-10
[SIL-89] Tests for reviewers screen logic

Krzysiudan:
https://silentus-au.atlassian.net/jira/software/projects/SIL/boards/1?selectedIssue=SIL-89

Changes: 
- added tests for `ReviewersViewModel`
- added MainDispatcherExtansion()
- removed the avatarUrl param from the `RequestedReviewer` data model. 

Note:
- I didn't write tests for error occurrences, I believe it will be easier to make it during error handling implementation.


app/src/test/java/com/appunite/loudius/ui/reviewers/ReviewersViewModelTest.kt
```diff
@@ -0,0 +1,125 @@
...
+        every { get<String>("pull_request_number") } returns "correctPullRequestNumber"
+    }
+    private lateinit var viewModel: ReviewersViewModel
+
+    private fun createViewModel() = ReviewersViewModel(repository, savedStateHandle)
+
+    @BeforeEach
+    fun setup() {
+        mockkStatic(Clock::class)
+        every { Clock.systemDefaultZone() } returns systemClockFixed
```
Krzysiudan, line 44: **info:** Needed to mock behaviour of the `LocalDateTime.now`.

app/src/test/java/com/appunite/loudius/ui/reviewers/ReviewersViewModelTest.kt
```diff
@@ -0,0 +1,125 @@
...
+    fun `GIVEN no values in saved state WHEN init THEN throw IllegalStateException`() {
+        every { savedStateHandle.get<String>(any()) } returns null
+
+        assertThrows<java.lang.IllegalStateException> {
+            createViewModel()
+        }
+    }
+
+    @Test
+    fun `GIVEN correct initial values WHEN init THEN state is correct`() {
```
kezc: suggestion: title 'state is correct' seems a little bit too broad, here you check only `pullRequestNumber`, while different fields are checked in different tests.

In addition, I feel that 3 of 4 `verify` could be moved to corresponding tests

app/src/test/java/com/appunite/loudius/fakes/FakePullRequestRepository.kt
```diff
@@ -0,0 +1,59 @@
...
+                ),
+            ),
+        )
+        "notExistingPullRequestNumber" -> Result.failure(WebException.UnknownError(404, null))
+        else -> Result.success(RequestedReviewersResponse(emptyList()))
+    }
+
+    override suspend fun getCurrentUserPullRequests(): Result<PullRequestsResponse> {
+        TODO("Not yet implemented")
+    }
```
nowakweronika, line 58: Question: This will be implemented in the next PR?

app/src/test/java/com/appunite/loudius/ui/reviewers/ReviewersViewModelTest.kt
Krzysiudan: Thanks, you are right. I've divided that into two tests. In the second I check only if all initial values are read once. I believe it is the most readable.

app/src/test/java/com/appunite/loudius/fakes/FakePullRequestRepository.kt
Krzysiudan: Hmm, it's not connected to the functionality I've implemented here so I didn't think about it. But yes it should have been tested already. Maybe implementing error handling will be a good time to write missing tests.

kezc: I'll implement it in my PR

kezc (approved):

nowakweronika (approved):
Lgtm

PR #33, opened 2023-03-14
[SIL-88] Exclude current user from the reviewers list

Krzysiudan:
https://silentus-au.atlassian.net/jira/software/projects/SIL/boards/1?selectedIssue=SIL-88

Changes: 
- current user reviews now not showing in the `ReviewersScreen`
- written test to the success cases to the `PullRequestRepository`. 

UI:

- [ ] Current user review is not showing up on the `ReviewersScreen` (It was happening only when the author commented on his PR)

Screenshot 2023-03-14 at 08 13 15


app/src/main/java/com/appunite/loudius/domain/PullRequestRepositoryImpl.kt
```diff
@@ -37,8 +38,18 @@ class PullRequestRepositoryImpl @Inject constructor(
         owner: String,
         repo: String,
         pullRequestNumber: String,
-    ): Result<List<Review>> =
-        pullRequestsNetworkDataSource.getReviews(owner, repo, pullRequestNumber)
+    ): Result<List<Review>> {
+        val currentUser = userDataSource.getUser()
+        return currentUser.flatMap { user ->
+            pullRequestsNetworkDataSource.getReviews(owner, repo, pullRequestNumber)
```
kezc: nitpick: those 2 tasks could be run in parallel (nitpick because this might require some additional work with coroutinescopes and I don't expect you to do it :D)

app/src/main/java/com/appunite/loudius/domain/PullRequestRepositoryImpl.kt
```diff
@@ -37,8 +38,18 @@ class PullRequestRepositoryImpl @Inject constructor(
...
+                .map { excludeCurrentUserReviews(it, user) }
+        }
+    }
+
+    private fun excludeCurrentUserReviews(
+        it: List<Review>,
+        user: User,
+    ) = it.filter { review -> review.user.id != user.id }
```
kezc: 1. nitpick: the function excludes the user passed as the argument, not the current one :P
2. suggestion: first parameter name could be better or extension function can be used

app/src/test/java/com/appunite/loudius/domain/PullRequestRepositoryImpTest.kt
```diff
@@ -0,0 +1,76 @@
...
+    private val userDataSource: UserDataSource = mockk {
+        coEvery { getUser() } returns Result.success(User(1, "user1"))
+    }
+    private val repository = PullRequestRepositoryImpl(pullRequestDataSource, userDataSource)
+
+    @Nested
+    inner class GetReviewsFunctionTest {
+
+        @Test
+        fun `GIVEN correct values WHEN get reviews THEN return result with reviews excluding ones from current user`() =
```
kezc, line 32: suggestion: `correct values` - I think it could be more meaningful :D Maybe correct pull request data

app/src/main/java/com/appunite/loudius/domain/PullRequestRepositoryImpl.kt
Krzysiudan: 1. That's a fact 👍 
2. Definitely `it` is not the best name for an param here xD
I transformed it into an extension function named `excludeUserReviews`.

app/src/main/java/com/appunite/loudius/domain/PullRequestRepositoryImpl.kt
Krzysiudan: Actually, I say to myself that I will try to do it and in the end, it wasn't so much extra work. Even I didn't need to correct the tests. Or I've done something wrong xD but it's working.

app/src/test/java/com/appunite/loudius/domain/PullRequestRepositoryImpTest.kt
Krzysiudan: But the `correct pull request data` is longer in 2 words 😅  And there are no other values than the pull request data, so it means the same anyway. `Correct values` == everything is okay at the inputs -> shorter and states that it's a happy path test. Happy like every android developer today.

app/src/main/java/com/appunite/loudius/domain/PullRequestRepositoryImpl.kt
```diff
@@ -38,18 +42,22 @@ class PullRequestRepositoryImpl @Inject constructor(
         owner: String,
         repo: String,
         pullRequestNumber: String,
-    ): Result<List<Review>> {
-        val currentUser = userDataSource.getUser()
-        return currentUser.flatMap { user ->
+    ): Result<List<Review>> = withContext(coroutineContext) {
```
kezc, line 45: praise: <3

kezc (approved):

PR #34, opened 2023-03-14
[SIL-68] Notifying mechanism

Krzysiudan:
https://silentus-au.atlassian.net/jira/software/projects/SIL/boards/1?selectedIssue=SIL-68

### **Changes:**
- added notifying mechanism
- added tests to every component with notify changes
- added showing snackbar on notify success

### **UI:**
Screenshot 2023-03-16 at 09 08 43

### **Checklist:**
- [ ] Ping some colleagues to make code review faster 😅 

### NOTE
- `ReviewersViewModelTest` diff got ugly. I wrapped the existing test into the `InitTest` inner class and unfortunately, it fogged target changes. So watch out, and don't waste your time on the first 130 lines.


kezc:
> @kezc

it works:D

app/src/test/java/com/appunite/loudius/ui/reviewers/ReviewersViewModelTest.kt
```diff
@@ -44,87 +45,113 @@ class ReviewersViewModelTest {
...
+
+                val expected = listOf(
+                    Reviewer(1, "user1", true, 7, 5),
+                    Reviewer(2, "user2", true, 7, 5),
+                )
+                val actual = viewModel.state.reviewers
+
+                assertEquals(expected, actual)
+            }
+    }
```
Krzysiudan, line 133: **Info:** You can skip that code, I've only wrapped this tests into inner class - `InitTest`

Krzysiudan:
@nowakweronika

app/src/main/java/com/appunite/loudius/network/model/request/NotifyRequestBody.kt
nowakweronika, line 1: Question: What do you think about `typealias` for body? Here is just one string in data class, do we need data class?

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
```diff
@@ -170,6 +216,12 @@ fun DetailsScreenPreview() {
     val reviewer4 = Reviewer(4, "Jacek", false, 24, 0)
     val reviewers = listOf(reviewer1, reviewer2, reviewer3, reviewer4)
     LoudiusTheme {
-        ReviewersScreenStateless(pullRequestNumber = "Pull request #1", reviewers = reviewers, {})
+        ReviewersScreenStateless(
+            pullRequestNumber = "1",
+            reviewers = reviewers,
+            {},
+            SnackbarHostState(),
+            {},
```
nowakweronika: Suggestion: I would make named params here :-))

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
```diff
@@ -97,6 +103,22 @@ class ReviewersViewModel @Inject constructor(
     private fun countHoursTillNow(submissionTime: LocalDateTime): Long =
         ChronoUnit.HOURS.between(submissionTime, LocalDateTime.now())
 
+    fun onAction(action: ReviewersAction) = when (action) {
+        is ReviewersAction.Notify -> notifyUser(action.userLogin)
+        is ReviewersAction.OnSnackbarDismiss -> state = state.copy(isSuccessSnackbarShown = false)
```
nowakweronika: Suggestion: I would make private method here, i.e. `dismissSnackbar()` with `state = state.copy(isSuccessSnackbarShown = false)`.

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
```diff
@@ -97,6 +103,22 @@ class ReviewersViewModel @Inject constructor(
...
+    private fun notifyUser(userLogin: String) {
+        val (owner, repo, pullRequestNumber) = getInitialValues(savedStateHandle)
+
+        viewModelScope.launch {
+            repository.notify(owner, repo, pullRequestNumber, "@$userLogin")
+                .onSuccess {
+                    state = state.copy(isSuccessSnackbarShown = true)
+                }
+        }
+    }
```
nowakweronika, line 120: Thought: Or we can make separate method for both functions with param, I think about something like this:
```
    fun onAction(action: ReviewersAction) = when (action) {
        is ReviewersAction.Notify -> notifyUser(action.userLogin)
        is ReviewersAction.OnSnackbarDismiss -> updateSnackbarVisibility(false)
    }

    private fun notifyUser(userLogin: String) {
        val (owner, repo, pullRequestNumber) = getInitialValues(savedStateHandle)

        viewModelScope.launch {
            repository.notify(owner, repo, pullRequestNumber, "@$userLogin")
                .onSuccess {
                    updateSnackbarVisibility(true)
                }
        }
    }

    private fun updateSnackbarVisibility(shouldShow: Boolean) {
        state = state.copy(isSuccessSnackbarShown = shouldShow)
    }
```

app/src/main/java/com/appunite/loudius/network/model/request/NotifyRequestBody.kt
Krzysiudan: It's about request to make it work. I tried different configurations and only this one works. If we add a type alias then probably nothing changes for the backend, it will still be a String - and only with `body` as a String, it doesn't work. 

The backend wants an object there. Otherwise, it returns `422` `InvalidRequest`. If you know other smart way how to do it then let me know ✌️

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
Krzysiudan: I see your point! Yes, using a separate method to update a single state parameter can be helpful when there's only one parameter being changed. However, often multiple parameters are changing, and then using state.copy(...) directly is the only way. Is this a proposition only for this situation or for every single state change? 🤔 

Regarding the additional states like isLoading and isError, if we would like to be consistent then we should have a function updating every part of the state, and I feel like it can not be so beneficial. 

Thanks for sharing your thoughts on this. It's always great to have different perspectives and ideas.

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
Krzysiudan: Added! 😄

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
Krzysiudan: Thanks! I like that 👌

app/src/main/java/com/appunite/loudius/network/model/request/NotifyRequestBody.kt
nowakweronika: Unfortunately I don't😅

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
nowakweronika: I thought about only this situation and I agree that it wouldn't be consistent to have just one function updating one part of the state. I think it's better to stay with what we have :D

nowakweronika (approved):
👌🏻

Krzysiudan:
@kezc

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
```diff
@@ -55,23 +90,28 @@ private fun ReviewersScreenStateless(
                 title = stringResource(id = R.string.details_title, pullRequestNumber),
             )
         },
+        snackbarHost = { SnackbarHost(hostState = snackbarHostState) },
         content = { padding ->
-            ReviewersScreenContent(reviewers, modifier = Modifier.padding(padding))
+            ReviewersScreenContent(reviewers, modifier = Modifier.padding(padding), onNotifyClick)
```
kezc: nitpick: all arguments could have been named, not only modifier

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
```diff
@@ -55,23 +90,28 @@ private fun ReviewersScreenStateless(
...
+) {
     LazyColumn(
         modifier = modifier.fillMaxWidth(),
     ) {
         itemsIndexed(reviewers) { index, reviewer ->
             ReviewerItem(
                 reviewer = reviewer,
                 backgroundColor = resolveReviewerBackgroundColor(index),
-                onNotifyClick = {},
+                onNotifyClick = onNotifyClick,
```
kezc, line 118: suggestion: probably not for this PR, but we should think about some instant feedback for the user because it takes a moment to show the snackbar and the user might think it doesn't work. Maybe we could disable the button until we show snackbar or something :)

kezc (approved):

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
Krzysiudan: Good idea, but I thought we would show the loading indicator after clicking on notify. So it will improve user experience. Do you agree with that? 🤔

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
Krzysiudan: Done that.

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
kezc: yes, it's even better idea

PR #37, opened 2023-03-17
[SIL-69] Reviewers screen error handling and loading during initial fetching data

Krzysiudan:
https://silentus-au.atlassian.net/jira/software/projects/SIL/boards/1?selectedIssue=SIL-69

### Changes:
- added showing error screen and loading on fetching data during initialization of the ReviewersScreen
- refactored fetching data on view model init to run in parallel
- added additional tests
- added default params for `LoudiusErrorScreen` and removed such from call places


app/src/test/java/com/appunite/loudius/fakes/FakePullRequestRepository.kt
```diff
@@ -2,58 +2,17 @@ package com.appunite.loudius.fakes
...
-                Review("2", User(1, "user1"), COMMENTED, date2),
-                Review("3", User(1, "user1"), APPROVED, date3),
-                Review("4", User(2, "user2"), COMMENTED, date1),
-                Review("5", User(2, "user2"), COMMENTED, date2),
-                Review("6", User(2, "user2"), APPROVED, date3),
-            ),
-        )
-        "notExistingPullRequestNumber" -> Result.failure(WebException.UnknownError(404, null))
-        else -> Result.success(emptyList())
-    }
```
Krzysiudan, line 38: **Info** : With the previous implementation I couldn't test the case that after retrying repository returns success response.

Krzysiudan:
@nowakweronika

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
```diff
@@ -55,14 +56,33 @@ class ReviewersViewModel @Inject constructor(
...
+                .onSuccess { state = state.copy(reviewers = it.orEmpty(), isLoading = false) }
+                .onFailure { state = state.copy(isError = true, isLoading = false) }
+        }
+    }
 
-        repository.getRequestedReviewers(owner, repo, pullRequestNumber)
-            .onSuccess { response ->
-                val reviewers = response.mapToReviewers(submissionTime)
-                state = state.copy(reviewers = state.reviewers + reviewers)
+    private suspend fun getMergedData(initialValues: InitialValues): Result<List<Reviewer>?> =
```
kezc: suggestion: do we need `initialValues` as a parameter here (and in other places) if it's the class's field? I think we can remove it

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
```diff
@@ -55,14 +56,33 @@ class ReviewersViewModel @Inject constructor(
...
+        coroutineScope {
```
kezc, line 68: suggestion: I don't think we need to create here a new coroutineScope. Is there anything against using `withContext(coroutineContext)`?

nowakweronika (approved):

Krzysiudan:
@kezc

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
Krzysiudan: Firstly I've used that because I thought it is simpler semantically than `withContext`. But I've done a little digging. They are very similar to each other. There is only a small difference.

> coroutineScope is meant to provide a scope for multiple parallel coroutines in which all will be cancelled, if any fails.

> withContext is designed to be used to switch the context (eg. the Dispatcher) for the given block of code.

And here I believe `coroutineScope` fits better. Previously when I used `withContext` I didn't know the difference. So I will change that `withContext` in the repository to the `coroutineScope` cause it's a similar situation.  

Here are the discussions if you want to:
https://stackoverflow.com/questions/56858624/kotlin-coroutines-what-is-the-difference-between-coroutinescope-and-withcontext
https://discuss.kotlinlang.org/t/coroutines-different-ways-to-change-coroutine-context/13095/3
https://stackoverflow.com/questions/56904057/why-does-withcontext-await-for-the-completion-of-child-coroutines

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
Krzysiudan: Good idea, done that 👌

PR #38, opened 2023-03-17
[SIL-93] Add loading indicator to login screen

nowakweronika:
### Description
*Add loading indicator to login screen.*

### Why?
*We want to have pretty UI :)*

### Demo
https://user-images.githubusercontent.com/72873966/225895964-29548253-0ca4-4717-b1de-854b70646b62.mov

### DIFF
--- a/app/src/main/java/com/appunite/loudius/ui/loading/LoadingScreen.kt
+++ b/app/src/main/java/com/appunite/loudius/ui/loading/LoadingScreen.kt
@@ -32,13 +32,13 @@ fun LoadingScreen(
             viewModel.onAction(LoadingAction.OnNavigateToPullRequests)
         }
     }
-    ResolveScreenToShow(showErrorScreen = state.showErrorScreen) {
+    LoadingScreenStateless(showErrorScreen = state.showErrorScreen) {
         viewModel.onAction(LoadingAction.OnTryAgainClick)
     }
 }
 
 @Composable
-fun ResolveScreenToShow(
+fun LoadingScreenStateless(
     showErrorScreen: Boolean,
     onTryAgainClick: () -> Unit,
 ) {
@@ -67,7 +67,7 @@ private fun ShowLoudiusErrorScreen(
 @Composable
 fun ShowLoudiusErrorScreenPreview() {
     LoudiusTheme {
-        ResolveScreenToShow(showErrorScreen = true) {}
+        LoadingScreenStateless(showErrorScreen = true) {}
     }
 }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/appunite/loudius/ui/loading/LoadingScreen.kt
```diff
@@ -54,9 +55,22 @@ private fun ShowLoudiusErrorScreen(
 }
 
 @Composable
-private fun ShowLoadingIndicator(code: String?) {
-    // TODO add loading indicator
-    Column {
-        Text(text = code ?: "code is already consumed")
+private fun ShowLoadingIndicator() {
+    LoudiusLoadingIndicator()
```
kezc: suggestion: I think we can remove this function

app/src/main/java/com/appunite/loudius/ui/loading/LoadingScreen.kt
```diff
@@ -37,7 +38,7 @@ fun LoadingScreen(
             viewModel.onAction(LoadingAction.OnTryAgainClick)
         }
     } else {
-        ShowLoadingIndicator(code = code)
+        ShowLoadingIndicator()
```
kezc: suggestion: what if we extracted this part (lines: 36-41) to another composable? It would allow us to do a proper preview where we would pass state, instead of displaying components

Krzysiudan (approved):
LGTM

app/src/main/java/com/appunite/loudius/ui/loading/LoadingScreen.kt
nowakweronika: Then, we should pass `viewModel` or function `onAction` too, do we want that?

kezc: pass `showErrorScreen` and `onAction`? I think it's the way we usually do it

app/src/main/java/com/appunite/loudius/ui/loading/LoadingScreen.kt
```diff
@@ -32,13 +32,22 @@ fun LoadingScreen(
...
         }
     }
+    ResolveScreenToShow(showErrorScreen = state.showErrorScreen) {
+        viewModel.onAction(LoadingAction.OnTryAgainClick)
+    }
+}
 
-    if (state.showErrorScreen) {
+@Composable
+fun ResolveScreenToShow(
```
kezc: nitpick: I thought it would be `LoadingScreenStateless`, because that's the way we usually make it. Maybe I should've taken today's action item more seriously.... 😬

kezc (approved):

PR #49, opened 2023-03-24
[LD-48] Add loading spinner upon clicking notify

kezc:
### Description
Added loading indicator which is shown upon clicking notify button

### Testing
Go to the reviewers screen and notify somebody:)

### DIFF
--- a/app/src/main/java/com/appunite/loudius/domain/model/Reviewer.kt
+++ b/app/src/main/java/com/appunite/loudius/domain/model/Reviewer.kt
@@ -6,4 +6,5 @@ data class Reviewer(
     val isReviewDone: Boolean,
     val hoursFromPRStart: Long,
     val hoursFromReviewDone: Long?,
+    val isLoading: Boolean = false,
 )
--- a/app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
+++ b/app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
@@ -33,7 +33,6 @@ import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import androidx.hilt.navigation.compose.hiltViewModel
 import com.appunite.loudius.R
-import com.appunite.loudius.domain.model.Reviewer
 import com.appunite.loudius.ui.components.LoudiusErrorScreen
 import com.appunite.loudius.ui.components.LoudiusLoadingIndicator
 import com.appunite.loudius.ui.components.LoudiusTopAppBar
--- a/app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
+++ b/app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
@@ -177,12 +177,12 @@ private fun ReviewerItem(
 @Composable
 private fun NotifyButtonOrLoadingIndicator(
     reviewer: Reviewer,
-    onNotifyClick: (ReviewersAction) -> Unit
+    onNotifyClick: (ReviewersAction) -> Unit,
 ) {
     val buttonAlpha = if (reviewer.isLoading) 0f else 1f
     Box(contentAlignment = Center) {
         NotifyButton(
-            modifier = Modifier.alpha(buttonAlpha)
+            modifier = Modifier.alpha(buttonAlpha),
         ) { onNotifyClick(ReviewersAction.Notify(reviewer.login)) }
 
         if (reviewer.isLoading) {
--- a/app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
+++ b/app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
@@ -2,12 +2,15 @@ package com.appunite.loudius.ui.reviewers
 
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.itemsIndexed
+import androidx.compose.material3.CircularProgressIndicator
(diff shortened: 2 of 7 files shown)


app/src/main/java/com/appunite/loudius/domain/model/Reviewer.kt
```diff
@@ -6,4 +6,5 @@ data class Reviewer(
     val isReviewDone: Boolean,
     val hoursFromPRStart: Long,
     val hoursFromReviewDone: Long?,
+    val isLoading: Boolean = false,
```
nowakweronika, line 9: Question: What do you think about putting this value into state in view model? I think, loading state is ui state and don't know if it's good practice to have this value here.

nowakweronika: But in the other hand, it's easy and clever solution, so I don't know what will be better :D

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
```diff
@@ -166,8 +171,22 @@ private fun ReviewerItem(
...
+}
+
+@Composable
+private fun NotifyButtonOrLoadingIndicator(
+    reviewer: Reviewer,
+    onNotifyClick: (ReviewersAction) -> Unit,
+) {
+    Box(contentAlignment = Center) {
+        NotifyButton(
+            modifier = Modifier.alpha(if (reviewer.isLoading) 0f else 1f),
```
Krzysiudan: **nitpick** I'm not a fan of inlining such if's. I would opt for extracting it to the variable.

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
```diff
@@ -166,8 +171,22 @@ private fun ReviewerItem(
...
+        ) { onNotifyClick(ReviewersAction.Notify(reviewer.login)) }
+
+        if (reviewer.isLoading) {
+            CircularProgressIndicator(modifier = Modifier.size(24.dp))
```
Krzysiudan: **question:** Why it's made like that? I mean why not to have it all in if clause? 
Like that: 

```
if (reviewer.isLoading) {
            CircularProgressIndicator(modifier = Modifier.size(24.dp))
        } else {
            NotifyButton(
                modifier = Modifier.alpha(if (reviewer.isLoading) 0f else 1f),
            ) { onNotifyClick(ReviewersAction.Notify(reviewer.login)) }
        }
```

There were some troubles with that way or you've made it because of something else? I'm curious.

app/src/test/java/com/appunite/loudius/ui/reviewers/ReviewersViewModelTest.kt
```diff
@@ -190,27 +191,51 @@ class ReviewersViewModelTest {
...
+
+            viewModel.onAction(ReviewersAction.Notify("user1"))
+
+            assertTrue(
+                viewModel.state.reviewers.first { it.login == "user1" }.isLoading,
+            ) { "Clicked item should have loading indicator" }
+            assertTrue(
+                viewModel.state.reviewers.filterNot { it.login == "user1" }.none { it.isLoading },
+            ) { "Only clicked item should have loading indicator" }
+        }
```
Krzysiudan: **Chore:** Update test name. There is no testing showing snackbar.

app/src/test/java/com/appunite/loudius/ui/reviewers/ReviewersViewModelTest.kt
```diff
@@ -190,27 +191,51 @@ class ReviewersViewModelTest {
...
         }
 
+        @Test
+        fun `WHEN successful notify action THEN show loading`() = runTest {
+            viewModel = createViewModel()
+
+            viewModel.onAction(ReviewersAction.Notify("user1"))
+
+            assertEquals(ReviewersSnackbarType.SUCCESS, viewModel.state.snackbarTypeShown)
+        }
```
Krzysiudan: **Chore: ** Update test name. Probably test names got mixed 😅

app/src/main/java/com/appunite/loudius/domain/model/Reviewer.kt
Krzysiudan: `Reviewer` model is a representation of each row, so I believe that it is a lot simpler to have the state of each row in one place. Moreover for me `Reviewer` is a lot like the UI model, we don't use it anywhere else than in the UI. So from my point of view, I would leave it as it is. 👌

Krzysiudan (approved):
LGTM.

app/src/main/java/com/appunite/loudius/domain/model/Reviewer.kt
nowakweronika: Okkk, nice solution. I was wondering if it's good practice but I trust you :D

nowakweronika (approved):

app/src/main/java/com/appunite/loudius/domain/model/Reviewer.kt
kezc: but I think we should move the `Reviewer` class to a different package because it's a UI model, not a domain model 🤔

Krzysiudan: Yeah sounds good to me, especially when we do the mapping in the view model and not somewhere else deeper in the hierarchy. 👌

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
kezc: I've had 2 problems:
- When the button has `alpha=0`, the size of the box stays the same. Placing the loading indicator inside this fixed-size box was the easiest way to center it within the button.
- Hiding the button sometimes would change the height of the whole item too.

app/src/test/java/com/appunite/loudius/ui/reviewers/ReviewersViewModelTest.kt
kezc: I don't know what happened there, because this test is copy paste from a different test :D thanks and already fixed:)

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
Krzysiudan: Ahh okay, sounds good for me then 👍

PR #51, opened 2023-03-27
[LD-52] Refactor dependency injections modules.

Krzysiudan:
https://loudius.atlassian.net/jira/software/projects/LD/boards/3

## Changes: 
- renamed modules to point to the type they provide 
- move around provides methods to be in correct module

As we agreed on the project structure to be more type-oriented I've refactored our DI modules to meet that requirement. 
Feel free to question mine refactor.


kezc (approved):
Good work, thank you for cleaning it

PR #52, opened 2023-03-27
[LD-52] Codebase improving - domain package

Krzysiudan:
https://loudius.atlassian.net/jira/software/projects/LD/boards/3

## Changes
- moved `AuthRepositoryImpl` into the file with `AuthRepository` interface
- moved repositories into the additional package - repository. 

Feel free to question mine refactor.


app/src/main/java/com/appunite/loudius/di/GithubModule.kt
```diff
@@ -1,9 +1,9 @@
 package com.appunite.loudius.di
```
kezc: thought: we've left a lonely `UserLocalDataSource` in the domain package. Maybe we could create a separate package for that? WDYT?

nowakweronika: Good idea :D

nowakweronika (commented):
conflict

app/src/main/java/com/appunite/loudius/di/GithubModule.kt
Krzysiudan: Yeah, good idea 👌 
Moved that to the `store` package.

nowakweronika (approved):

kezc (approved):

PR #57, opened 2023-03-30
[LD-56/LD-20]/ Handling auth failure.

Krzysiudan:
https://loudius.atlassian.net/jira/software/projects/LD/boards/3?selectedIssue=LD-56
https://loudius.atlassian.net/jira/software/projects/LD/boards/3?selectedIssue=LD-20

## Changes: 
- added handling out auth failure in the interceptor
- added navigating to the login screen and showing toast on auth failure
- added tests for interceptor and `MainViewModel`. 

I've changed the implementation of emitting to the flow. During pair programming, we had global flow. Unfortunately, it's impossible to unit test with the global flow.  That is why I've created a separate component. Additionally, I've moved that from the `safeApiCall` method because otherwise, I would have to rebuild/move a lot of tests. 

## Testing 
While logged in, revoke authorization for Loudius at your GitHub profile settings, and then perform some action, for example, try to notify the reviewer. 
Screenshot 2023-03-30 at 12 36 43


app/src/main/java/com/appunite/loudius/ui/MainViewModel.kt
```diff
@@ -0,0 +1,35 @@
+package com.appunite.loudius.ui
```
kezc, line 17: issue: license

app/src/main/java/com/appunite/loudius/network/utils/AuthFailureHandler.kt
```diff
@@ -0,0 +1,23 @@
+package com.appunite.loudius.network.utils
```
kezc, line 17: issue: license

app/src/main/res/values/strings.xml
```diff
@@ -16,9 +16,9 @@
     <string name="try_again">Try again</string>
     <string name="error_login_text">Something went wrong…\nYou need to log in again.</string>
     <string name="go_to_login">Take me to login</string>
-    <string name="reviewers_snackbar_message">Awesome! Your collaborator have been pinged for some serious code review action! 🎉</string>
     <string name="reviewers_snackbar_success">Awesome! Your collaborator have been pinged for some serious code review action! 🎉</string>
     <string name="reviewers_snackbar_failure">Uh-oh, it seems that Loudius has taken a vacation. Don\'t worry, we\'re sending a postcard to bring it back ASAP!</string>
+    <string name="user_unauthorized_message">"Unauthorized collaborator detected! Please login again."</string>
```
kezc: issue: unnecessary parentheses 
```suggestion
    <string name="user_unauthorized_message">Unauthorized collaborator detected! Please login again.</string>
```

app/src/main/java/com/appunite/loudius/network/utils/AuthFailureHandler.kt
```diff
@@ -0,0 +1,23 @@
...
+import javax.inject.Singleton
+
+interface AuthFailureHandler {
+    suspend fun emitAuthFailure()
+
+    val authFailureFlow: SharedFlow<Unit>
+}
+
+@Singleton
+class AuthFailureHandlerImpl @Inject constructor() : AuthFailureHandler {
```
kezc: suggestion: you provide this class through Hilt's `@Provide` so it's redundant
```suggestion
class AuthFailureHandlerImpl : AuthFailureHandler {
```

app/src/main/java/com/appunite/loudius/ui/MainViewModel.kt
```diff
@@ -0,0 +1,35 @@
...
+    init {
+        viewModelScope.launch {
+            authFailureHandler.authFailureFlow.collect {
+                state = MainState(Unit)
+            }
+        }
+    }
+
+    fun onAuthFailureHandled() {
+        state = MainState(null)
```
kezc: nitpick: null is the default value, so it could be
```suggestion
        state = MainState()
```

app/src/main/java/com/appunite/loudius/MainActivity.kt
```diff
@@ -99,4 +111,22 @@ class MainActivity : ComponentActivity() {
...
     }
+
+    private fun navigateToLoginOnAuthFailure(navController: NavHostController) {
+        if (viewModel.state.authFailureEvent != null) {
+            showAuthFailureToast()
+            navController.navigate(Screen.Login.route) {
+                popUpTo(navController.graph.id) { inclusive = true }
+            }
+            viewModel.onAuthFailureHandled()
+        }
```
nowakweronika, line 122: Question: What about phones with a back button at the bottom of the screen? Maybe we should do a `popBackStack` here.

app/src/main/java/com/appunite/loudius/ui/MainViewModel.kt
nowakweronika, line 1: praise: I like it :D

app/src/main/java/com/appunite/loudius/network/intercept/AuthFailureInterceptor.kt
```diff
@@ -0,0 +1,42 @@
...
+
+class AuthFailureInterceptor @Inject constructor(
+    private val authFailureHandler: AuthFailureHandler,
+) : Interceptor {
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val request = chain.request()
+        val response = chain.proceed(request)
+
+        if (response.code == 401) {
+            CoroutineScope(Dispatchers.IO).launch {
```
nowakweronika: suggestion: I think, we should inject a dispatcher instead of hardcoding it. 
Here is article about it on the android developers website https://developer.android.com/kotlin/coroutines/coroutines-best-practices#inject-dispatchers

nowakweronika (commented):
Question: If we are on the collaborators screen, revoke the app and click back, we are moved to the PRs screen, do we want this behaviour?

app/src/main/java/com/appunite/loudius/MainActivity.kt
Krzysiudan: `popUpTo` is doing the same, it is clearing the backstack because I've passed graph id as a parameter - thanks to that every destination is removed.

app/src/main/java/com/appunite/loudius/network/intercept/AuthFailureInterceptor.kt
kezc: actually, should it be on the IO dispatcher? It doesn't perform any IO operations

Krzysiudan: I can do that, but I don't feel like it is much of an improvement because if its purpose is for making the test easier then our tests are already easy. I bet it's more usable with more complicated tasks or I don't see some problem with the code now. 

> All TestDispatchers should share the same scheduler. This allows you to run all your coroutine code on a single test thread to make your tests deterministic. runTest will wait for all coroutines that are on the same scheduler or are children of the test coroutine to complete before returning.

So now in tests, if we want to test our AuthFailureInterceptor we need to create `okHttpClient` at the beginning of the test to give the dispatcher the correct scheduler. 

So every test should start with the: 
```
val testOkHttpClient = testOkHttpClient(authFailureHandler = fakeAuthFailureHandler, dispatcher = UnconfinedTestDispatcher(testScheduler))
            val service = retrofitTestDouble(mockWebServer = mockWebServer, client = testOkHttpClient,).create(TestApi::class.java)
```
            
Maybe I don't understand something but if our tests work anyway then what's the difference? Please share some opinions with me.

nowakweronika: Instead of easier testing, there is one more advantage. If we wanted to add another dispatcher, it would be easier by passing an argument. Anyway, I don't know more advantages and I don't know if they are sufficient to make this changes :)

Krzysiudan:
> Question: If we are on the collaborators screen, revoke the app and click back, we are moved to the PRs screen, do we want this behaviour?

I believe it isn't a big issue. If the user revoked the permission a while ago then he have seen that information anyway. 
But refreshing the PR's list on moving back could be a better experience. So maybe it's worth spending time with.

app/src/main/java/com/appunite/loudius/network/intercept/AuthFailureInterceptor.kt
Krzysiudan: @kezc  Nicely spotted. I've changed the dispatcher to `Default`
@nowakweronika I've added the dispatcher in the arguments. I believe we can try to use the proper dispatcher in the test later. 👌

nowakweronika (approved):

kezc (approved):

PR #59, opened 2023-04-03
Jacek's review WIP

jacek-marchwicki:
# Why?
Because this is an example project, unlike normal project, code is more important than feature set. I'd did want to review the code and make some improvements that would improve readibility of the code.

# What?
I added a lot of suggestions about code. Those are my personal opinions, so feel free to present your arguments and opinions about my suggestions. I'm not alpha and omega.
This code contains mainly suggestions, not solution, so the code probably won't build.

PR #60, opened 2023-04-04
Cleanup authenticating screen

jacek-marchwicki:
# Why?
I'd like to give some advises how to write clean code. I'd like to apply some suggestions listed here: https://github.com/appunite/Loudius/pull/59

# What:
Mainly, I'd like to change the loading screen.

* Change "Loading" screen to "Authenticating" screen
* Handle passing intent directly to view model by SavedStateHandle instead of passing it from activity, by ui to viewmodel. 
* Some cleanup in test names
* Using spy in tests
* Better handle case where code is not passed to activity (`adb shell am start -a android.intent.action.VIEW -d "loudius://callback"`)


app/src/main/java/com/appunite/loudius/common/Screen.kt
```diff
@@ -16,16 +16,33 @@
...
 
-    object Repos : Screen("repos_screen")
+    object Authenticating : Screen("repos_screen") {
+        fun getCode(savedStateHandle: SavedStateHandle): Result<String> {
+            val intent: Intent? = savedStateHandle[NavController.KEY_DEEP_LINK_INTENT]
+            val code = intent?.data?.getQueryParameter("code")
+            return code?.let { Result.success(it) } ?: Result.failure(Exception("No error code"))
+        }
+
+        val deepLinks: List<NavDeepLink> get() = listOf(
```
kezc, line 40: suggestion: I think the getter is not needed here, it could be a regular variable

kezc (approved):

app/src/main/java/com/appunite/loudius/common/Screen.kt
jacek-marchwicki: It could, but I had problem with tests where in tests it can't create NavBuildLinks

PR #63, opened 2023-04-04
[LD-60] Add missing tests in PullRequestRepositoryImpTest

kezc:
### Description
Added missing test for failure path and getting current user pull requests

### DIFF
--- a/app/src/test/java/com/appunite/loudius/domain/PullRequestRepositoryImpTest.kt
+++ b/app/src/test/java/com/appunite/loudius/domain/PullRequestRepositoryImpTest.kt
@@ -42,7 +42,7 @@ class PullRequestRepositoryImpTest {
 
     private val pullRequestDataSource = spyk(FakePullRequestDataSource())
     private val userDataSource: UserDataSource = mockk {
-        coEvery { getUser() } returns Result.success(User(1, "user1"))
+        coEvery { getUser() } returns Result.success(Defaults.currentUser())
     }
     private val repository = PullRequestRepositoryImpl(pullRequestDataSource, userDataSource)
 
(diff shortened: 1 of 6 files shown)


app/src/test/java/com/appunite/loudius/domain/PullRequestRepositoryImpTest.kt
```diff
@@ -46,35 +50,48 @@ class PullRequestRepositoryImpTest {
     inner class GetReviewsFunctionTest {
 
         @Test
-        fun `GIVEN correct values WHEN get reviews THEN return result with reviews excluding ones from current user`() =
+        fun `GIVEN correct pull request data WHEN getting reviews THEN return result with reviews excluding ones from current user`() =
             runTest {
-                val actual = repository.getReviews("example", "example", "correctPullRequestNumber")
+                val pullRequestNumber = "correctPullRequestNumber"
```
jacek-marchwicki, line 54: question: why do you extract this variable: `val pullRequestNumber = "correctPullRequestNumber"`  ithink it makes harder to understand this test.

app/src/test/java/com/appunite/loudius/domain/PullRequestRepositoryImpTest.kt
```diff
@@ -85,28 +102,98 @@ class PullRequestRepositoryImpTest {
...
                 "@ExampleUser",
             )
 
-            val expected = Result.success(Unit)
+            assertEquals(Result.success(Unit), actual)
+        }
 
-            assertEquals(expected, actual)
+        @Test
+        fun `GIVEN incorrect pull request number WHEN notifying THEN return success result`() = runTest {
```
jacek-marchwicki: chore: oops `incorrect pull request` -> `return success result`

app/src/test/java/com/appunite/loudius/domain/PullRequestRepositoryImpTest.kt
```diff
@@ -85,28 +102,98 @@ class PullRequestRepositoryImpTest {
...
         }
-        // TODO: Write tests with failure cases
     }
+
+    @Nested
+    inner class GetPullRequestsForUserTest {
+        @Test
+        fun `GIVEN logged in user WHEN getting user's pull requests THEN return pull requests`() =
+            runTest {
+                coEvery { userDataSource.getUser() } returns Result.success(
```
jacek-marchwicki, line 169: suggestion: I think by default the mock recurs correct data. You don't need to mock it.

jacek-marchwicki (approved):
lgtm

app/src/test/java/com/appunite/loudius/domain/PullRequestRepositoryImpTest.kt
kezc: I did it because I thought an explicit 'GIVEN' part would more clearly show what is the difference between the tests

app/src/test/java/com/appunite/loudius/domain/PullRequestRepositoryImpTest.kt
kezc: I have added a big commit now, where I introduce to tests `currentUser`. I've talked to Krzysztof and we think that it might get hard to control many different fakes/make them work together if we keep using different names for the current user, e.g. `user1`, `exampleUser`.

Starting from now I would suggest setting the username of the currently logged-in user as `currentUser`, so it's consistent across the tests. If you agree, I will add it to the working agreement and make a message on Slack.

PR #64, opened 2023-04-04
[LD-79] Improve styles

jacek-marchwicki:
# Why?
In the example project, it would be great using Design System pattern. This PR presents how to use this pattern in the code.

# What?
Extract common styles to composables. Add a lot of previews ;) 
* Use everywhere `LoudiusOutlinedButton` instead of `OutlinedButton` to have exactly the same styles
* Use `LoudiusText` instead of `Text` so we limit number of styles.
* Extract common `LoudiusListItem` from PullRequestsScreen and `ReviewersScreen`

Keep in mind that styles might differ from what was implemented before.


app/src/main/java/com/appunite/loudius/ui/components/LoudiusListItem.kt
```diff
@@ -0,0 +1,205 @@
...
+
+@Composable
+fun resolveListItemBackgroundColor(index: Int): Color =
+    if (index % 2 == 0) MaterialTheme.colorScheme.onSurface.copy(0.08f) else MaterialTheme.colorScheme.surface
+
+@Composable
+fun LoudiusListItem(
+    index: Int,
+    modifier: Modifier = Modifier,
+    icon: @Composable (Modifier) -> Unit = {},
```
kezc, line 46: thought:
Sometimes we use `icon: @Composable (Modifier) -> Unit = {},` (e.g. here) and sometimes we use `icon: @Composable (() -> Unit)? = null` (`LoudiusOutlinedButton`), maybe we should stick to one convention. I don't know which one is better because null is used in the compose library (check `InputChip`), but empty lambda seems to be a cleaner solution, at least for me.

app/src/main/java/com/appunite/loudius/ui/components/LoudiusOutlinedButton.kt
```diff
@@ -16,42 +16,139 @@
...
-            color = MaterialTheme.colorScheme.tertiary,
+            modifier = Modifier
+                .applyIf<Modifier>(style == LoudiusOutlinedButtonStyle.Large) { padding(8.dp) }
+                .applyIf(style == LoudiusOutlinedButtonStyle.Regular && icon != null) { padding(start = 8.dp) },
+            style = LoudiusTextStyle.Button,
+        )
+    }
+}
+
+inline fun <T> T.applyIf(predicate: Boolean, block: T.() -> T): T {
```
kezc, line 64: suggestion: I would extract it to another file as it seems to be useful and is a public function

kezc (approved):

app/src/main/res/drawable/ic_pull_request.xml
```diff
@@ -1,9 +1,12 @@
...
-</vector>
+      android:pathData="m6.193,4.147a1.487,1.487 0,1 0,-2.231 1.287l0,4.862a1.487,1.487 0,1 0,1.487 0L5.449,5.434A1.487,1.487 0,0 0,6.193 4.147ZM4.705,3.404A0.744,0.744 0,1 1,3.962 4.147,0.744 0.744,0 0,1 4.705,3.404ZM4.705,12.327a0.744,0.744 0,1 1,0.744 -0.744,0.744 0.744,0 0,1 -0.744,0.744z"
+      android:fillColor="#000000"/>
+  <path
+      android:pathData="M11.398,10.296L11.398,5.542A2.14,2.14 135,0 0,9.26 3.404L8.795,3.404L8.795,2.66A0.372,0.372 0,0 0,8.161 2.397l-1.487,1.487a0.372,0.372 0,0 0,0 0.526l1.487,1.487A0.372,0.372 0,0 0,8.795 5.634l0,-0.744l0.465,0a0.651,0.651 0,0 1,0.651 0.651l0,4.755a1.487,1.487 0,1 0,1.487 0zM10.655,12.327a0.744,0.744 0,1 1,0.744 -0.744,0.744 0.744,0 0,1 -0.744,0.744z"
+      android:fillColor="#000000"/>
+</vector>
```
Krzysiudan, line 12: **Chore** No new line at the end of the file.

app/src/main/java/com/appunite/loudius/ui/components/LoudiusListItem.kt
jacek-marchwicki: I use `(Modifer)`, when it should be passed to children so it can apply the modifier. The same as Scaffold pass `padding` only in `content` parameter

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -122,46 +120,43 @@ private fun PullRequestsList(
...
-    Image(
+private fun PullRequestIcon(modifier: Modifier) {
+    LoudiusListIcon(
+        modifier = modifier,
         painter = painterResource(id = R.drawable.ic_pull_request),
-        contentDescription = null,
-        modifier = Modifier
-            .padding(start = 18.dp, top = 10.dp)
-            .size(width = 18.dp, height = 20.dp),
+        contentDescription = stringResource(id = R.string.pull_requests_screen_pull_request_content_description),
```
Krzysiudan, line 146: **Question:** Is "Pull Requests" the correct content description for an icon? 
Also, the string name probably could be shorter - like `pull_request_icon_content_description`.

app/src/main/java/com/appunite/loudius/ui/components/LoudiusOutlinedButton.kt
jacek-marchwicki: It could be moved, but I do not like to make `common` every function that were useful in one place, this makes common a place for storing one use methods ;) But if there will be a place where the function could be used again. I'm totally fine moving this function out of this file :)

Krzysiudan (approved):
**Praise:** Well done, it seems better way to do it 👌

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
jacek-marchwicki: I've added a task for this: https://loudius.atlassian.net/browse/LD-82

app/src/main/res/drawable/ic_pull_request.xml
jacek-marchwicki: Ooops... next time ;)

PR #71, opened 2023-04-07
[LD-75] Create simple test for checking login screen

kezc:
### Description
Added a sample test for checking if the login screen opens.
### Why
- We want to test UI and check if the login screen launches correctly 
- We want to have any UI test to configure CI
#### Testing
Run androidTest tests


app/build.gradle
```diff
@@ -102,9 +106,11 @@ dependencies {
     implementation 'com.google.code.gson:gson:2.10.1'
 
     // assertion library
-    testImplementation 'io.strikt:strikt-core:0.34.0'
-    testImplementation 'io.strikt:strikt-mockk:0.34.0'
-    androidTestImplementation 'io.strikt:strikt-core:0.34.0'
+    // cannot use 0.34.0 due to an existing bug
+    // https://github.com/robfletcher/strikt/issues/259
```
jacek-marchwicki, line 110: praise: nice documentation ❤️

app/src/androidTest/java/com/appunite/loudius/LoginScreenTest.kt
```diff
@@ -0,0 +1,46 @@
+package com.appunite.loudius
```
jacek-marchwicki, line 17: chore: no license header in all files

app/src/debug/java/com/appunite/loudius/TestActivity.kt
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package com.appunite.loudius
+
+import androidx.activity.ComponentActivity
+import dagger.hilt.android.AndroidEntryPoint
+
+@AndroidEntryPoint
+class TestActivity : ComponentActivity()
```
kezc, line 23: note: This is in debug folder so we don't clutter the production build. I don't know if we want to run the tests in debug or production flavor, but if in the latter, we will have to change it.

app/src/androidTest/java/com/appunite/loudius/LoginScreenTest.kt
```diff
@@ -0,0 +1,57 @@
...
+    @get:Rule(order = 1)
+    val composeTestRule = createAndroidComposeRule<TestActivity>()
+
+    @Before
+    fun setUp() {
+        hiltRule.inject()
+    }
+
+    @Test
+    fun whenTheLoginScreenIsVisibleThenTheLogInButtonIsVisible() {
```
Krzysiudan, line 48: **Question:** In instrumented tests it's not possible to use convention with "`"?

Krzysiudan (approved):
lgtm

app/src/androidTest/java/com/appunite/loudius/LoginScreenTest.kt
kezc: unfortunately it's not possible

app/src/androidTest/java/com/appunite/loudius/LoginScreenTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+@RunWith(AndroidJUnit4::class)
+@HiltAndroidTest
+class LoginScreenTest {
+
+    @get:Rule(order = 0)
+    val hiltRule = HiltAndroidRule(this)
+
+    @get:Rule(order = 1)
+    val composeTestRule = createAndroidComposeRule<TestActivity>()
```
jacek-marchwicki, line 40: chore: if you'll use `val composeTestRule = createComposeRule()` instead of `createAndroidComposeRule<TestActivity>()` then you'll not need `TestActivity`, so you can skip changes in the AndroidManifest.xml file.

kezc: This is true as long as this screen does not have a view model. This version of the code does not include your changes to check if a device is xiaomi, but I checked it on a test branch where I merged it with your code and it wasn't working.

( similar problem https://stackoverflow.com/questions/71344878/real-viewmodel-with-composable-injected-by-hilt-in-instrumentation-test-zero-arg?noredirect=1&lq=1 )

app/src/androidTest/java/com/appunite/loudius/LoginScreenTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    @Before
+    fun setUp() {
+        hiltRule.inject()
+    }
```
kezc, line 45: suggestion: this actually is unnecessary

PR #72, opened 2023-04-11
[LD-83] Safe api call method improvements. 

Krzysiudan:
https://loudius.atlassian.net/jira/software/projects/LD/boards/3?selectedIssue=LD-83

## Changes: 
- wrapped `safeApiCall` method into class `ApiRequester` to reuse Gson instance
- added catching `JsonParseException` in `safeApiCall`
- extracted `BadVerificationCodeException` from `WebException` sealed class
- improved `LocalDateTimeDeserializer`.

## Issue fixed:

Screenshot 2023-04-11 at 13 43 02
image
image


jacek-marchwicki (approved):
lgtm

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
```diff
@@ -100,4 +101,7 @@ object NetworkModule {
         GsonBuilder().setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
             .registerTypeAdapter(LocalDateTime::class.java, LocalDateTimeDeserializer())
             .create()
+
+    @Provides
```
kezc: suggestion: it's redundant, because you have `@Inject` annotation in `ApiRequester` and it doesn't implement any interfaces

kezc (approved):

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
Krzysiudan: Thanks for that 👌  Firstly I got the error with not resolved deps so I've added provides.

PR #73, opened 2023-04-12
[LD-72] Integration tests CI

kezc:
### Description
- Added workflow to run UI tests on Firebase Test Lab. Currently, we only run tests on one emulated device - Pixel 2
- Extracted 2 actions
    - `prepare-android-env` for preparing jdk and gradle
    - `clean-up-gradle-cache` for clearing gradle cache

As you can see in this PR, the test passes correctly. Here https://github.com/appunite/Loudius/actions/runs/4698047629/jobs/8329849106?pr=73 you can see a non-passing test.


Krzysiudan (approved):
LGTM!

.github/actions/clean-up-gradle-cache/action.yml
Krzysiudan, line 1: **Praise:** Nice extraction of that 2 actions 👏

.github/actions/clean-up-gradle-cache/action.yml
```diff
@@ -0,0 +1,10 @@
+name: "Clean up gradle cache workflow"
+description: "Remove some files from the Gradle cache, so they aren't cached by GitHub Actions. Restoring these files from a GitHub Actions cache might cause problems for future builds."
```
jacek-marchwicki: question: I know this was done before, but I have a few questions to the code:
1. Why do you think, this might cause problems for the future builds? If the problems happen, is there any other way of just cleaning the case in the situation?
2. You've had set up caching by using `gradle/gradle-build-action@v2`, why you wont to disable it now?
3. There is an option to disable cache in gradle-build-action plugin: https://github.com/gradle/gradle-build-action#caching, why not use this option?

.github/workflows/run-ui-test.yml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Prepare Android Environment
+        uses: ./.github/actions/prepare-android-env
+
+      - name: Assemble app debug APK
```
jacek-marchwicki: suggestion: when you'll execute `assembleDebug` and `assembleDebugAndroidTest` in one command, you can better leverage multicore, while those two commands are executed together. I think about: `./gradlew assembleDebug assembleDebugAndroidTest`

jacek-marchwicki (approved):

.github/actions/clean-up-gradle-cache/action.yml
kezc: 1. We have it because I got it from Wire's CI setup... I think we could remove it for now, see if the problems occur, and eventually restore it. What do you think?
2. I don't want to disable it, but I see it's both in `prepare-android-env` and `run-unit-test`
3. I didn't know about this option.

jacek-marchwicki: | I think we could remove it for now, see if the problems occur, and eventually restore it. What do you think?
Good! ;)

jacek-marchwicki (approved):
lgtm

PR #74, opened 2023-04-17
[LD-84] Auth failure improvements

Krzysiudan:
## Changes: 
- made AuthFailureInterceptor not use the coroutine
- corrected CoroutineHelpers package 
- remove obsolete AuthFailureHandler interface. 

## Why
To improve the code by fixing issues described here: 
https://loudius.atlassian.net/jira/software/projects/LD/boards/3?selectedIssue=LD-84

### DIFF
--- a/app/src/main/java/com/appunite/loudius/di/GeneralModule.kt
+++ b/app/src/main/java/com/appunite/loudius/di/GeneralModule.kt
@@ -14,24 +14,19 @@
  * limitations under the License.
  */
 
-package com.appunite.loudius.network.intercept
+package com.appunite.loudius.di
 
-import com.appunite.loudius.network.utils.AuthFailureHandler
-import javax.inject.Inject
-import okhttp3.Interceptor
-import okhttp3.Response
+import dagger.Module
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
 
-class AuthFailureInterceptor @Inject constructor(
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/appunite/loudius/di/GeneralModule.kt
```diff
@@ -0,0 +1,32 @@
...
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+
+@InstallIn(SingletonComponent::class)
+@Module
+object GeneralModule {
+
+    @Provides
+    fun provideDefaultDispatcher(): CoroutineDispatcher = Dispatchers.Default
+}
```
Krzysiudan: **Info** This can feel like a redundant piece of code, but it's needed because the Hilt has a problem with the default value of the argument if there is only one argument. 

It is widely described in the following StackOverflow post : 
https://stackoverflow.com/questions/55240472/types-may-only-contain-one-inject-constructor

app/src/main/java/com/appunite/loudius/di/GeneralModule.kt
```diff
@@ -0,0 +1,32 @@
...
+import dagger.Module
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+
+@InstallIn(SingletonComponent::class)
+@Module
+object GeneralModule {
```
kezc: suggestion: Maybe it could be named `DispatchersModule`?

kezc (approved):

nowakweronika (approved):

PR #75, opened 2023-04-17
[LD-86] Use fake shared preferences for UserLocalDataSourceTest instead of mock

Krzysiudan:
## Changes: 
- used fake shared preferences for UserLocalDataSourceTest instead of mock

## Why: 
To improve the code proposed by suggestion in #59


kezc (approved):
LGTM

nowakweronika (approved):

PR #76, opened 2023-04-17
[LD-85] View state in sealed class

Krzysiudan:
## Changes: 
- refactored PullRequestScreen to use view state in sealed class

## Why: 
To improve the code by the suggestion proposed in #59 
image

https://loudius.atlassian.net/jira/software/projects/LD/boards/3/backlog?issueParent=10117&selectedIssue=LD-85

## Reason for a draft: 
I'm not so convinced that there are more gains than pains from refactoring to the view state in sealed class. Moreover, in implementation with a sealed class every time we would like to get the data from the `Loaded` state we need to cast it, so that's extra work to do. 
Can you share your opinions with me? 
Or maybe you have an idea of how to upgrade my implementation? 🤔

### DIFF
--- a/app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
+++ b/app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
@@ -69,7 +69,7 @@ fun PullRequestsScreen(
 private fun navigateToReviewers(
     state: PullRequestState,
     navigateToReviewers: NavigateToReviewers,
-    viewModel: PullRequestsViewModel
+    viewModel: PullRequestsViewModel,
 ) {
     state.navigateToReviewers?.let {
         navigateToReviewers(
@@ -100,7 +100,6 @@ private fun PullRequestsScreenStateless(
                 is Data.Loading -> LoudiusLoadingIndicator(Modifier.padding(padding))
                 is Data.Loaded -> PullRequestContent(state.data, padding, onAction)
             }
-
         },
     )
 }
@@ -234,7 +233,7 @@ fun PullRequestsScreenPreview() {
                             createdAt = LocalDateTime.parse("2022-01-29T16:31:41"),
                         ),
                     ),
-                )
+                ),
             ),
             onAction = {},
         )
--- a/app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
+++ b/app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
@@ -97,10 +97,10 @@ private fun PullRequestsScreenStateless(
                     modifier = Modifier.padding(padding),
                     onButtonClick = { onAction(PulLRequestsAction.RetryClick) },
                 )
-
                 is Data.Loading -> LoudiusLoadingIndicator(Modifier.padding(padding))
                 is Data.Loaded -> PullRequestContent(state.data, padding, onAction)
             }
+
         },
     )
 }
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -57,50 +57,67 @@ fun PullRequestsScreen(
...
+                    LaunchedEffect(state.navigateToReviewers) {
+                        state.navigateToReviewers?.let {
+                            navigateToReviewers(
+                                it.owner,
+                                it.repo,
+                                it.pullRequestNumber,
+                                it.submissionTime
+                            )
+                            onAction(PulLRequestsAction.OnNavigateToReviewers)
+                        }
```
Krzysiudan: **Info:** In that case it would be probably better to move navigation events to a separate flow.

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsViewModel.kt
```diff
@@ -33,12 +33,14 @@ sealed class PulLRequestsAction {
...
-    val isLoading: Boolean = false,
-    val isError: Boolean = false,
-)
+sealed class PullRequestState {
+    object Loading : PullRequestState()
+    object Error : PullRequestState()
+    data class Loaded(
+        val pullRequests: List<PullRequest> = emptyList(),
+        val navigateToReviewers: NavigationPayload? = null,
+    ) : PullRequestState()
```
nowakweronika: Suggestion: What do you think about `Success` instead of `Loaded` or something like that?

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -57,50 +57,67 @@ fun PullRequestsScreen(
...
+                        state.navigateToReviewers?.let {
+                            navigateToReviewers(
+                                it.owner,
+                                it.repo,
+                                it.pullRequestNumber,
+                                it.submissionTime,
+                            )
+                            onAction(PulLRequestsAction.OnNavigateToReviewers)
+                        }
+                    }
```
nowakweronika: Suggestion: Maybe we can extract this part of code as a function? WDYT?

nowakweronika (commented):
Overall I like this concept more that previous :D

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsViewModel.kt
jacek-marchwicki: suggestion: What do you think about:

```kotlin
sealed class Data {
  object Loading : Data()
  object Error : PullRequestState()
  data class Loaded(val pullRequests: List<PullRequest>)
}

data class PullRequestState(
  val data: Data,
  val navigateToReviewers: NavigationPayload? = null,
)
```

So `navigateToReviewers` is not related to the sealed class.

jacek-marchwicki (approved):
lgtm

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
Krzysiudan: Done

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsViewModel.kt
Krzysiudan: Done with that extra `Data` class.

PR #78, opened 2023-04-18
[LD-64]: Make UI presentable on all devices screen

nowakweronika:
## Description
A lot of our screens doesn’t work on smaller screens. We need to ensure everything is presented fine on smaller screens too. I checked the bigger one too :) Give me feedback please, WDYT about bigger screen? Should we implement different UI for them?

## Link
[LD-64 Ensure app is working on smaller screens](https://loudius.atlassian.net/browse/LD-64)

## Demo
| 4" | 5.6" | 10.1" |
|------------------|----------------|----------------|
| Screenshot 2023-04-19 at 12 21 16 | Screenshot 2023-04-19 at 12 21 36 | Screenshot 2023-04-19 at 12 22 48 |

### DIFF
--- a/app/src/main/java/com/appunite/loudius/ui/components/LoudiusFullScreenError.kt
+++ b/app/src/main/java/com/appunite/loudius/ui/components/LoudiusFullScreenError.kt
@@ -19,17 +19,23 @@ package com.appunite.loudius.ui.components
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalConfiguration
+import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import com.appunite.loudius.R
 import com.appunite.loudius.ui.theme.LoudiusTheme
+
+private const val minHeightToFitSpacer = 1856
 
 @Composable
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/appunite/loudius/ui/components/LoudiusFullScreenError.kt
```diff
@@ -19,31 +19,92 @@ package com.appunite.loudius.ui.components
...
 @Composable
 fun LoudiusFullScreenError(
     modifier: Modifier = Modifier,
     errorText: String = stringResource(id = R.string.error_dialog_text),
     buttonText: String = stringResource(id = R.string.try_again),
     onButtonClick: () -> Unit,
+) {
+    val density = LocalDensity.current
+    val configuration = LocalConfiguration.current
+    val screenHeight = with(density) { configuration.screenHeightDp.dp.roundToPx() }
```
kezc, line 48: question: have you considered comparing values in dp instead of px? I think it could be a safer solution. You can see the sample sizes here https://developer.android.com/guide/topics/large-screens/support-different-screen-sizes#window_size_classes

app/src/main/java/com/appunite/loudius/ui/components/LoudiusFullScreenError.kt
```diff
@@ -19,31 +19,92 @@ package com.appunite.loudius.ui.components
...
+                .fillMaxWidth(),
+            horizontalAlignment = Alignment.CenterHorizontally,
+            verticalArrangement = Arrangement.SpaceBetween,
+        ) {
+            ErrorImage()
+            ErrorText(text = errorText)
+            LoudiusOutlinedButton(
+                onClick = onButtonClick,
+                text = buttonText,
+            )
```
Krzysiudan, line 90: **Suggestion** Those 3 composables can be extracted to single function and then reused in `ScreenErrorWithoutSpacers` and `ScreenErrorWithSpacers`.

Krzysiudan (approved):
LGTM.

app/src/main/java/com/appunite/loudius/ui/components/LoudiusFullScreenError.kt
nowakweronika: I can't take values from this article because the smallest screen is in `medium height`. I can only do it in dp instead of px, but with this values.

kezc (approved):

nowakweronika:
@kezc

PR #80, opened 2023-04-20
Nicer way of scaling

jacek-marchwicki:
**WARNING**: Consider this PR as PoC. I haven't verified this on device, so before merging, please test this change.

# Why?
We would like to have a screen that scales nicely across different devices

# What?

While implementing screens that should be presented nicely on different screen sizes, there is a natural tendency to distinguish devices between different categories: screens that are larger than X, smaller than Y. This might be good for some cases, but usually we should try not to implement two different screens, but make screen behave correctly on all screens sizes. Especially, we should keep in mind, that devices might be portrait and landscape. 

For testing this in jetpack compose, I suggest using previews with different screen sizes, that helps ensure that the screen looks great for a variety of devices, without a need for running the app on all those devices type at once. 
It's a good practice to specify those common devices types in some constant values, so we don't need to think too much about those devices types every time we implement a new screen.

Moreover, if we're not creating completely different UI for different screen sizes, it is better to make some small adjustments to the screen then creating completely new screen from scratch. This will make our app less prone to errors like forgetting to add a button only on one screen size.
Those adjustments might be:
* Change the font size depending on the screen size
* Set bigger vertical padding on device wider than 620dp.
* Add weighted parameter as padding (column/row with spacers)

On screens, where we mainly present short amount of data, like welcome screens, error screens. It's good practice to use weighed parameters. It's also good practice to add left and right paddings.

# Screenshots
| | | | | | |
|-|-|-|-|-|-|
| img | | img_1 | img_2 | img_3 | img_4 |

# TODO
- [x] During Code Review, I verified the PoC implementation.


app/src/main/java/com/appunite/loudius/ui/components/LoudiusFullScreenError.kt
```diff
@@ -155,3 +104,51 @@ fun LoudiusErrorScreenPreview() {
...
+            errorText = stringResource(id = R.string.error_dialog_text),
+            buttonText = stringResource(R.string.try_again),
+            onButtonClick = {},
+        )
+    }
+}
+
+@Preview(showSystemUi = true, device = ReferenceDevices.SmallPhoneLandscape)
+@Composable
+fun LoudiusErrorScreenPreviewSmallPhoneLandscape() {
```
marcin-adamczewski: AFAIK they recently added multi-preview annotation. I think it could work here well.
[LINK](https://developer.android.com/jetpack/compose/tooling/previews#preview-multipreview)

Krzysiudan (approved):
Lgtm! 
Thanks for sharing, that's a nice way of scaling 👌

nowakweronika (approved):

jacek-marchwicki:
@marcin-adamczewski

PR #81, opened 2023-04-28
[LD-102] Screen recording with paparazzi

Krzysiudan:
## Why?
To show an example implementation of golden testing using the paparazzi library.

## Changes: 
- Created new module `components` 
This is the requirement of the paparazzi library. For now, it's impossible to use paparazzi in the applications module, library module is needed. 
- Moved all UI components to the separate module `components` with resources used by these components
- Added git LFS 
It is recommended because screenshots can take up a lot of memory

## How to check

- [x] Use `./gradlew screenshots:rPD` (`./gradlew screenshots:recordPaparazziDebug`) and open generated report in a browser|
You should see something like that: 
Screenshot 2023-05-15 at 16 09 12

- [x] Delete one of the button previews in the `LoudiusButtonTests` for example `LoudiusOutlinedButtonPreview()` and use `./gradlew screenshots:vPD` (`./gradlew screenshots:verifyPaparazziDebug`). 
This should result in failing tests and 2 new images like that: 

Screenshot 2023-05-15 at 16 13 45
Screenshot 2023-05-15 at 16 14 02

- [x] Play around and do some nice picks like paparazzi

## Warning 
Before committing any screenshots to the repo please install git LFS : 
https://docs.github.com/en/repositories/working-with-files/managing-large-files/installing-git-large-file-storage


app/src/main/java/com/appunite/loudius/MainActivity.kt
```diff
@@ -31,12 +31,12 @@ import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.composable
 import androidx.navigation.compose.rememberNavController
 import com.appunite.loudius.common.Screen
+import com.appunite.loudius.screenshots.theme.LoudiusTheme
```
kezc: nitpick: is screenshot a good name for a module with components? :D I would name it `components`

kezc (approved):
LGTM!

app/src/main/java/com/appunite/loudius/MainActivity.kt
Krzysiudan: The name is problematic a little for now. `componets` would be really good if we move every component to this module. But right now there are only a few of them. Maybe it's a good idea to move all components. 🤔

screenshots/src/main/java/com/appunite/loudius/screenshots/theme/Theme.kt
```diff
@@ -69,13 +69,14 @@ fun LoudiusTheme(
...
         }
+
         darkTheme -> DarkColorScheme
         else -> LightColorScheme
     }
     val view = LocalView.current
     if (!view.isInEditMode) {
         SideEffect {
-            (view.context as Activity).window.statusBarColor = colorScheme.primary.toArgb()
+            (view.context as? Activity)?.window?.statusBarColor = colorScheme.primary.toArgb()
```
nowakweronika: Question: Why do we need to make the safe call?

nowakweronika (approved):
lgtm :)

app/src/main/java/com/appunite/loudius/ui/authenticating/AuthenticatingScreen.kt
```diff
@@ -22,9 +22,10 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.hilt.navigation.compose.hiltViewModel
 import com.appunite.loudius.R
-import com.appunite.loudius.ui.components.LoudiusFullScreenError
+import com.appunite.loudius.screenshots.R.string
```
jacek-marchwicki: I'd suggest instead of importing `R.string` I'd use:
```
import com.appunite.loudius.screenshots.R as screenshotsR`
```

app/src/main/java/com/appunite/loudius/MainActivity.kt
jacek-marchwicki: I don't like the `screenshot` package name, either. I think moving all common components is a good approach. If you prefer, you can do this in a followup change.

screenshots/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+    buildFeatures {
+        compose true
+    }
+    composeOptions {
+        kotlinCompilerExtensionVersion '1.4.2'
+    }
+}
+
+dependencies {
+    def composeBom = platform('androidx.compose:compose-bom:2023.01.00')
```
jacek-marchwicki: question: why this is extracted to the variable (`composeBom`)?

screenshots/src/main/java/com/appunite/loudius/screenshots/components/LoudiusFullScreenError.kt
```diff
@@ -56,11 +59,17 @@ fun ScreenErrorWithSpacers(
     onButtonClick: () -> Unit,
 ) {
     Column(
-        modifier = modifier.padding(32.dp).fillMaxSize(),
+        modifier = modifier
+            .background(color = Color.White)
```
jacek-marchwicki: question: why you have you had to add background modifier?

screenshots/src/test/java/com/appunite/loudius/screenshots/LoudiusButtonTests.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    @get:Rule
+    val paparazzi = Paparazzi(
+        deviceConfig = DeviceConfig.PIXEL_5,
+        renderingMode = SessionParams.RenderingMode.V_SCROLL,
+        showSystemUi = false,
+    )
+
+    @Test
+    fun loudiusOutlinedButton() {
```
jacek-marchwicki: question: wouldn't be better to have a different test for all of those views?

screenshots/src/test/snapshots/images/com.appunite.loudius.screenshots_LoudiusButtonTests_loudiusOutlinedButton.png
jacek-marchwicki, line 1: question: why those buttons are not aligned in the same way?

screenshots/build.gradle
jacek-marchwicki, line 1: question: there is no way, to use `app` module. Do we need to create new one?

screenshots/src/main/java/com/appunite/loudius/screenshots/theme/Theme.kt
Krzysiudan: Good question 👌 
The problem (actually it's more like a feature) is that paparazzi not launching views in edit mode. Because of that, the build is failing - Context is type `BridgeContext` instead of `Activity`.

screenshots/build.gradle
jacek-marchwicki: I guess it's not possible: https://github.com/cashapp/paparazzi/issues/107 :)

app/build.gradle
```diff
@@ -64,6 +64,8 @@ android {
 }
 
 dependencies {
+    implementation project(":screenshots")
```
jacek-marchwicki: question: If screenshot doesn't match, will CI detect the issue?

screenshots/build.gradle
Krzysiudan: Not needed, copying leftover. Thanks for spotting 👍

app/src/main/java/com/appunite/loudius/ui/authenticating/AuthenticatingScreen.kt
Krzysiudan: Good idea 👌

screenshots/src/main/java/com/appunite/loudius/screenshots/components/LoudiusFullScreenError.kt
Krzysiudan: I will provide a link to a document(work in progress): 
https://www.notion.so/appunite/Experiment-Golden-tests-for-jetpack-compose-74e68882a08f4126b9b1d76d22f55706?pvs=4#5eb9d312a0f2435eb4d9241a1a976820

Basically on the screenshot this `LoudiusFullScreenError()` had no background before. So it looked bad. 
I couldn't find the reason why it is like that.

screenshots/src/main/java/com/appunite/loudius/screenshots/theme/Theme.kt
nowakweronika: Nice, thanks for the answer 😼

screenshots/src/test/java/com/appunite/loudius/screenshots/LoudiusButtonTests.kt
Krzysiudan: Do you think it would be better? 
Every test it's a whole device screen size so I thought it would be an overkill to have screenshots of every single component (plus later we could have snapshots for different configurations).

screenshots/src/test/snapshots/images/com.appunite.loudius.screenshots_LoudiusButtonTests_loudiusOutlinedButton.png
Krzysiudan: Because of the applyIf we have in `LoudiusOutlinedButton()` for Large buttons. It renders the same in AS preview. 
```kotlin
OutlinedButton(
        enabled = enabled,
        onClick = onClick,
        modifier = modifier
            .applyIf(style == LoudiusOutlinedButtonStyle.Large) { padding(horizontal = 46.dp) },
    ) {
```

screenshots/build.gradle
Krzysiudan: Unfortunately for now it's not :/

app/build.gradle
Krzysiudan: Yes it should not be the problem. 
I've seen one example GitHub action. At the end of this article https://betterprogramming.pub/sanely-test-your-android-ui-libraries-with-paparazzi-b6d46c55f6b0
Screenshot 2023-05-16 at 13 31 07

jacek-marchwicki: Is this already implemented? Have you verified those failures?

screenshots/src/test/snapshots/images/com.appunite.loudius.screenshots_LoudiusButtonTests_loudiusOutlinedButton.png
jacek-marchwicki: Ops... looks like you found some issue inside the code... This shouldn't work like this ;) Margins of components should be added outside the component

app/build.gradle
Krzysiudan: No, I didn't work on CI yet so I didn't verify how this failure works. But I will take care of CI soon.

screenshots/src/test/snapshots/images/com.appunite.loudius.screenshots_LoudiusButtonTests_loudiusOutlinedButton.png
Krzysiudan: Yeah, you're right 😅 I will create a small PR with a fix for that. 
Or maybe I will leave it as a change for verifying snapshots, and then I will fix that.

app/build.gradle
jacek-marchwicki: Np. You can do this later, but IMHO it's a pretty important part of the research.

app/src/main/java/com/appunite/loudius/MainActivity.kt
Krzysiudan: I've changed the module name to the `components` here. ✅

components/build.gradle
```diff
@@ -0,0 +1,46 @@
...
+    }
+    kotlinOptions {
+        jvmTarget = '11'
+    }
+    buildFeatures {
+        compose true
+    }
+    composeOptions {
+        kotlinCompilerExtensionVersion '1.4.2'
+    }
```
jacek-marchwicki, line 29: out-of-scope/suggestion: I think it would be great to add resource prefix to ensure all components resources. So we're sure what resources are common. We could add prefix requirements in the build.gradle file:
```
resourcePrefix 'components_'
```

jacek-marchwicki (approved):
lgtm

components/build.gradle
Krzysiudan: Sounds good 👌 
I've added this to the current PR. Lines corresponding to resources were changed anyway.

PR #83, opened 2023-05-02
Example how we could potentiall use OkHttp mockwebserver

jacek-marchwicki:


app/src/androidTest/java/com/appunite/loudius/util/MockWebServerRule.kt
```diff
@@ -0,0 +1,93 @@
+package com.appunite.loudius.util
```
jacek-marchwicki, line 1: chore: missing license

app/src/main/java/com/appunite/loudius/di/TestInterceptor.kt
```diff
@@ -0,0 +1,16 @@
+package com.appunite.loudius.di
```
jacek-marchwicki, line 1: chore: missing license

PR #84, opened 2023-05-08
[LD-68] Mock web server

nowakweronika:
## Description
*Add a mock web server to mock API responses in integration tests.*

### Why?
*We want to check if our code works properly.*

## Links to related issues

- [LD-68 Setup ability to mock API](https://loudius.atlassian.net/browse/LD-68)


app/src/androidTest/java/com/appunite/loudius/util/MockWebServerRule.kt
```diff
@@ -0,0 +1,93 @@
...
+                MockWebServer().use { server ->
+                    server.dispatcher = dispatcher
+                    TestInterceptor.testInterceptor = UrlOverrideInterceptor(server.url("/"))
+                    Log.v(TAG, "TestInterceptor installed")
+                    try {
+                    } finally {
+                        base.evaluate()
+                        Log.v(TAG, "TestInterceptor uninstalled")
+                        TestInterceptor.testInterceptor = null
+                    }
```
kezc: question: why is there an empty try block? Is it intended?

app/src/androidTest/java/com/appunite/loudius/PullRequestsScreenTest.kt
```diff
@@ -0,0 +1,159 @@
...
+        hiltRule.inject()
+        composeTestRule.setContent {
+            LoudiusTheme {
+                PullRequestsScreen { _, _, _, _ -> }
+            }
+        }
+    }
+
+    @Test
+    fun whenTheLoginScreenIsVisibleThenTheLogInButtonIsVisible() {
```
kezc: chore: the name is copied from a different place :D

app/src/androidTest/java/com/appunite/loudius/util/MockWebServerRule.kt
nowakweronika: We cannot just put a finally block here without try. We want to ensure that `TestsInterceptor` has been uninstalled and that the `TestInterceptor.testInterceptor` field is set to null.
@jacek-marchwicki correct me if I'm wrong :D

kezc: I don't think `finally` is executed if the exceptions occur outside try/catch block.  Take a look at my examples
Screenshot 2023-05-09 at 12 43 26
Screenshot 2023-05-09 at 12 43 44

jacek-marchwicki: This should be:
```kotlin
try {
  base.evaluate()
} finally {
  Log.v(TAG, "TestInterceptor uninstalled")
  TestInterceptor.testInterceptor = null
}
```

instead of:

```kotlin
try {
} finally {
  base.evaluate()
  Log.v(TAG, "TestInterceptor uninstalled")
  TestInterceptor.testInterceptor = null
}
```

app/src/androidTest/java/com/appunite/loudius/PullRequestsScreenTest.kt
```diff
@@ -0,0 +1,159 @@
...
+                    ]
+                }
+        """.trimIndent()
+
+        every {
+            mockWebServer.dispatcher.dispatch(matchArg { path.startsWith("/search/issues") })
+        } returns jsonResponse(jsonResponse)
+
+        sleep(3000) // Temporary solution
+        composeTestRule.onNodeWithTag(TestTags.PULL_REQUEST_ITEM).assertIsDisplayed()
```
jacek-marchwicki: suggestion: while doing end-to-end tests, I prefer testing from the user perspective. You wouldn't ask someone to check if a view with a tag is displayed. I think we would rather test if the given PR is displayed. You can simply do this by string check. 
I'd suggest changing `"title":"example title",` to something more meaningful, i.e. `"title":"First Pull-Request title", then checking: if there is a view with "First Pull-Request title" title.

app/build.gradle
```diff
@@ -50,6 +50,15 @@ android {
     packagingOptions {
         resources {
             excludes += '/META-INF/{AL2.0,LGPL2.1}'
+            excludes += '/META-INF/LICENSE.md'
+            excludes += '/META-INF/LICENSE-notice.md'
+        }
+    }
+    testOptions {
+        packagingOptions {
```
jacek-marchwicki, line 58: question: is this necessary? If so, could you describe why?

app/src/androidTest/java/com/appunite/loudius/util/MockWebServerRule.kt
```diff
@@ -0,0 +1,93 @@
...
+private const val TAG = "MockWebServerRule"
+
+class MockWebServerRule : TestRule {
+
+    val dispatcher: Dispatcher = mockk {
+        every { shutdown() } returns Unit
+        every { peek() } returns MockResponse().apply { this.socketPolicy = SocketPolicy.KEEP_OPEN }
+        every { dispatch(any()) } answers {
+            val request = it.invocation.args[0] as RecordedRequest
+            Log.w(TAG, "Request is not mocked: ${request.method} ${request.path}")
```
jacek-marchwicki, line 49: suggestion/out-of-scope: I found some tool that I've written before for generating Curls for given requests: https://github.com/jacek-marchwicki/blog/pull/2/files#diff-a3868ac818c91fa781d8971ef674100cb8902486617f216532ff273888289851R109-R152. This might be useful to call real method.

app/src/main/java/com/appunite/loudius/di/TestInterceptor.kt
```diff
@@ -0,0 +1,16 @@
+package com.appunite.loudius.di
```
jacek-marchwicki, line 17: chore: missing license header

app/src/androidTest/java/com/appunite/loudius/util/MockWebServerRule.kt
```diff
@@ -0,0 +1,93 @@
+package com.appunite.loudius.util
```
jacek-marchwicki, line 17: chore: missing license header

jacek-marchwicki (changes requested):

app/build.gradle
nowakweronika: Yes, it is necessary. Without this code, the error `MockK could not self-attach a jvmti agent to the current VM.` occurs. We need this code to ensure that the packaging mechanism is compatible with all devices we use for testing.

kezc (approved):

nowakweronika:
@jacek-marchwicki

jacek-marchwicki (approved):
lgtm

jacek-marchwicki (approved):
lgtm

PR #89, opened 2023-05-17
[LD-70] Add test for clicking login button

kezc:
## Description
I've created a test checking if a correct intent is created upon clicking the login button
### Why?
Because we want to cover our app with UI tests and logging in is a crucial feature of the app.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -131,7 +131,7 @@ dependencies {
     androidTestImplementation("com.squareup.okhttp3:mockwebserver:4.11.0")
     androidTestImplementation("io.mockk:mockk-android:1.13.3")
 
-    androidTestImplementation 'androidx.test.espresso:espresso-intents:3.4.0'
+    androidTestImplementation 'androidx.test.espresso:espresso-intents:3.5.1'
 
     // ktlint
     ktlintRuleset project(":custom-ktlint-rules")
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -131,6 +131,8 @@ dependencies {
     androidTestImplementation("com.squareup.okhttp3:mockwebserver:4.11.0")
     androidTestImplementation("io.mockk:mockk-android:1.13.3")
 
+    androidTestImplementation 'androidx.test.espresso:espresso-intents:3.4.0'
+
     // ktlint
     ktlintRuleset project(":custom-ktlint-rules")
 }
--- a/app/src/androidTest/java/com/appunite/loudius/LoginScreenTest.kt
+++ b/app/src/androidTest/java/com/appunite/loudius/LoginScreenTest.kt
@@ -16,18 +16,28 @@
 
 package com.appunite.loudius
 
-import androidx.compose.ui.test.assertIsDisplayed
+import android.app.Activity
+import android.app.Instrumentation
+import android.content.Intent
 import androidx.compose.ui.test.junit4.createAndroidComposeRule
 import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
+import androidx.test.espresso.intent.Intents
+import androidx.test.espresso.intent.Intents.intended
+import androidx.test.espresso.intent.Intents.intending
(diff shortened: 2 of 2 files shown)


app/src/androidTest/java/com/appunite/loudius/LoginScreenTest.kt
```diff
@@ -46,12 +55,24 @@ class LoginScreenTest {
...
             }
         }
 
-        composeTestRule.onNodeWithText("Log in").assertIsDisplayed()
+        composeTestRule.onNodeWithText("Log in").performClick()
+
+        intended(
+            allOf(
+                hasAction(Intent.ACTION_VIEW),
+                hasData("https://github.com/login/oauth/authorize?client_id=91131449e417c7e29912&scope=repo"),
```
nowakweronika, line 73: Question: Do we want pass `client_id` like that? Or maybe we can pass whole url as value? I think something like `private const val AUTH_URL = "https://github.com/login/oauth/authorize?client_id=91131449e417c7e29912&scope=repo"` would be more readable.

app/build.gradle
```diff
@@ -131,6 +131,8 @@ dependencies {
     androidTestImplementation("com.squareup.okhttp3:mockwebserver:4.11.0")
     androidTestImplementation("io.mockk:mockk-android:1.13.3")
 
+    androidTestImplementation 'androidx.test.espresso:espresso-intents:3.4.0'
```
nowakweronika: Question: Why 3.4.0? There is newer version - 3.5.1.

app/src/androidTest/java/com/appunite/loudius/LoginScreenTest.kt
nowakweronika, line 75: Question: Do we want pass `client_id` like that? Or maybe we can pass whole url as value? I think something like `private const val AUTH_URL = "https://github.com/login/oauth/authorize?client_id=91131449e417c7e29912&scope=repo"` would be more readable.

app/src/androidTest/java/com/appunite/loudius/LoginScreenTest.kt
```diff
@@ -46,12 +55,24 @@ class LoginScreenTest {
 
     @Test
     fun whenTheLoginScreenIsVisibleThenTheLogInButtonIsVisible() {
+        Intents.init()
```
nowakweronika: Suggestion/Question: I'm wondering if `@Test` is the best place for `Intents.init()`. Maybe we should put it in `@Before` function in case we need it in another test.

kezc: If I needed this in a different test I would probably do it, but currently, I don't have any other test. Also, I am not sure if it affects tests in any way if we don't use `intended/intending`, probably no, but I don't want to mess up new tests with it unintentionally

app/src/androidTest/java/com/appunite/loudius/LoginScreenTest.kt
kezc: If it were an app's code I would totally do it... But in the tests, I care about the result and want exactly this URL, not any other... If i had to test it manually I would see this URL in Chrome too. I just like it more this way : - (

nowakweronika (approved):

Krzysiudan (approved):
LGTM! Nice, small and clean 👌

app/src/androidTest/java/com/appunite/loudius/LoginScreenTest.kt
jacek-marchwicki: suggestion: I suggest using `IntentsRule`:
```
@get:Rule(order = 1)
val intentsRule = IntentsRule()
```

Especially because in case of a test failure, `Intents.release()` is going to be released. 

It shouldn't be a problem if intents is not used in a given test. It should'n make any unnecessary action.

jacek-marchwicki (approved):

PR #91, opened 2023-05-18
[LD-80] Test for notify screen

nowakweronika:
## Description
*Test for checking if reviewers are visible when response is correct.*

### Why?
*We want to be sure that reviewers are visible in the correct way.*

### DIFF
--- a/app/src/androidTest/java/com/appunite/loudius/PullRequestsScreenTest.kt
+++ b/app/src/androidTest/java/com/appunite/loudius/PullRequestsScreenTest.kt
@@ -21,18 +21,13 @@ import androidx.compose.ui.test.onNodeWithText
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.appunite.loudius.ui.pullrequests.PullRequestsScreen
 import com.appunite.loudius.ui.theme.LoudiusTheme
+import com.appunite.loudius.util.Register
 import com.appunite.loudius.util.TestRules
-import com.appunite.loudius.util.jsonResponse
-import com.appunite.loudius.util.path
-import com.appunite.loudius.util.queryParameter
-import com.appunite.loudius.util.url
 import dagger.hilt.android.testing.HiltAndroidTest
 import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
-import strikt.api.expectThat
-import strikt.assertions.isEqualTo
 
 @RunWith(AndroidJUnit4::class)
 @HiltAndroidTest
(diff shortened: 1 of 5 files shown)


app/src/androidTest/java/com/appunite/loudius/ReviewersScreenTest.kt
```diff
@@ -0,0 +1,71 @@
...
+    val testRules = TestRules(this)
+
+    @Before
+    fun setUp() {
+        testRules.setUp()
+    }
+
+    @Test
+    fun whenResponseIsCorrectThenReviewersAreVisible() {
+        with(testRules) {
```
kezc: nitpick: if we use `=` then we can save one indentation 
```suggestion
     fun whenResponseIsCorrectThenReviewersAreVisible() = with(testRules) {
```

kezc (approved):

app/src/androidTest/java/com/appunite/loudius/ReviewersScreenTest.kt
```diff
@@ -0,0 +1,71 @@
...
+                reviews(mockWebServer)
+            }
+
+            composeTestRule.setContent {
+                LoudiusTheme {
+                    ReviewersScreen { }
+                }
+            }
+
+            composeTestRule.onRoot()
```
kezc: suggestion: Should it be here?

app/src/androidTest/java/com/appunite/loudius/PullRequestsScreenTest.kt
```diff
@@ -17,154 +17,45 @@
...
-
-        mockWebServer.register {
-            expectThat(it).url.and {
-                get("host") { host }.isEqualTo("api.github.com")
-                path.isEqualTo("/search/issues")
-                queryParameter("q").isEqualTo("author:jacek type:pr state:open")
-                queryParameter("page").isEqualTo("0")
-                queryParameter("per_page").isEqualTo("100")
+        with(testRules) {
+            with(Register) {
```
jacek-marchwicki: opinion: IMHO this trick is a little bit hacky and hard to understand what's going on ;)
I think much easier to understand would be:
```kotlin
mockWebServer.register(Responses.getUser)
mockWebServer.register(Responses.searchIssues)
```

app/src/androidTest/java/com/appunite/loudius/util/TestRules.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.appunite.loudius.util.IdlingResourceExtensions.toIdlingResource
+import dagger.hilt.android.testing.HiltAndroidRule
+import dagger.hilt.android.testing.HiltAndroidTest
+import org.junit.Before
+import org.junit.rules.RuleChain
+import org.junit.rules.TestRule
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+
+@HiltAndroidTest
```
jacek-marchwicki: question: I think `@HiltAndroidTest` is not necessary there, because it's not a test

app/src/androidTest/java/com/appunite/loudius/util/TestRules.kt
```diff
@@ -0,0 +1,50 @@
...
+class TestRules(testClass: Any) : TestRule {
```
jacek-marchwicki: suggestin: name `TestRules` is a little bit too generic, ;) maybe `IntegrationTestRule`

app/src/androidTest/java/com/appunite/loudius/util/TestRules.kt
```diff
@@ -0,0 +1,50 @@
...
+    private val hiltRule = HiltAndroidRule(testClass)
+
+    override fun apply(base: Statement, description: Description): Statement {
+        return RuleChain.outerRule(mockWebServer)
+            .around(hiltRule)
+            .around(composeTestRule)
+            .apply(base, description)
+    }
+
+    @Before
```
jacek-marchwicki: chore: this annotation would work only if this is executed as test, but not inside others tests

app/src/androidTest/java/com/appunite/loudius/util/TestRules.kt
```diff
@@ -0,0 +1,50 @@
...
+    fun setUp() {
+        composeTestRule.registerIdlingResource(countingResource.toIdlingResource())
```
jacek-marchwicki: suggestion: You can try move this registerIdlingResource to constructor
```
val composeTestRule = createAndroidComposeRule<TestActivity>().apply {
        registerIdlingResource(countingResource.toIdlingResource())
    }
    ```

File: app/src/androidTest/java/com/appunite/loudius/ReviewersScreenTest.kt (hunk 1 above)
Comment (nowakweronika) [reply app/src/androidTest/java/com/appunite/loudius/ReviewersScreenTest.kt]: I can't do this because I've got this error.
```
Invalid test class 'com.appunite.loudius.ReviewersScreenTest':
1. Method whenResponseIsCorrectThenReviewersAreVisible() should be void
```

File: app/src/androidTest/java/com/appunite/loudius/ReviewersScreenTest.kt (hunk 2 above)
Comment (nowakweronika) [reply app/src/androidTest/java/com/appunite/loudius/ReviewersScreenTest.kt]: No, we don't need that :-)

File: app/src/androidTest/java/com/appunite/loudius/PullRequestsScreenTest.kt [hunk 8]
```diff
@@ -17,154 +17,43 @@
...
-    @get:Rule(order = 1)
-    val hiltRule = HiltAndroidRule(this)
-
-    @get:Rule(order = 0)
-    var mockWebServer: MockWebServerRule = MockWebServerRule()
-
-    @get:Rule(order = 2)
-    val composeTestRule = createAndroidComposeRule<TestActivity>()
+    @get:Rule
+    val integrationTestRule = IntegrationTestRule(this)
```
Krzysiudan, line 37: **Praise:** That's nice extraction. I didn't know if that's possible 👌

app/src/androidTest/java/com/appunite/loudius/util/Register.kt
Krzysiudan, line 1: **Praise: ** Awesome idea with moving responses here. 👌  I'm not so convinced if `Register` is the best name for that object but anyway I like that.

Krzysiudan (approved):
LGTM

app/src/androidTest/java/com/appunite/loudius/PullRequestsScreenTest.kt
nowakweronika: It's thanks to you : -)

app/src/androidTest/java/com/appunite/loudius/util/Register.kt
nowakweronika: Idea was to use it like this `Register.user()` witch exactly describes what it's doing, but if someone has better name I will consider it :D

jacek-marchwicki: IMHO we could do this better, and I'm open for suggestions in upcoming PR's. But the proposed solution is also great, and much, much better of what we've had before.

jacek-marchwicki (approved):
lgtm

kezc (approved):
lgtm!

PR #92, opened 2023-05-22
[LD-80] Tests for notify screen

nowakweronika:
### Description
*Added tests for notifying user.*

#### Why?
*We want to ensure that we notify user in the proper way.*


Krzysiudan (approved):
Looks good to me!

Krzysiudan (approved):
Looks good to me!

kezc (approved):

PR #94, opened 2023-05-24
[LD-90] Add pull to refresh

nowakweronika:
## Description
*Added pull to refresh feature to ReviewersScreen and PullRequestsScreen.*


app/build.gradle
```diff
@@ -77,6 +77,7 @@ dependencies {
     implementation composeBom
     androidTestImplementation composeBom
     implementation 'androidx.compose.material3:material3'
+    implementation 'androidx.compose.material:material:1.4.0'
```
nowakweronika, line 80: Material Design 3 doesn't support pull to refresh mechanism, we have to wait for it :(

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
```diff
@@ -74,7 +74,7 @@ class ReviewersViewModel @Inject constructor(
         fetchData()
     }
 
-    private fun fetchData() {
+    fun fetchData() {
```
jacek-marchwicki: The same as above

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsViewModel.kt
```diff
@@ -62,7 +62,7 @@ class PullRequestsViewModel @Inject constructor(
         fetchData()
     }
 
-    private fun fetchData() {
+    fun fetchData() {
```
jacek-marchwicki: question: if you re-used the same method for fetching isn't a problem, where data disappears when user does pull-to-refresh? Usually when user doesn't pull-to-refresh, the old data should be still presented.

suggestion: If you make this class open, it would be beneficial adding unit tests for this class.

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -57,8 +63,13 @@ fun PullRequestsScreen(
     navigateToReviewers: NavigateToReviewers,
 ) {
     val state = viewModel.state
+    val swipeRefreshState = rememberPullRefreshState(
+        refreshing = state.data == Data.Loading,
+        onRefresh = { viewModel.fetchData() },
+    )
     PullRequestsScreenStateless(
         state = state,
+        pullRefreshState = swipeRefreshState,
```
jacek-marchwicki: suggestion: maybe we will stick to `pullRefreshState` name instead of `swipeRefreshState` what do you think?

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
```diff
@@ -61,10 +69,15 @@ fun ReviewersScreen(
...
     val snackbarHostState = remember { SnackbarHostState() }
+    val swipeRefreshState = rememberPullRefreshState(
+        refreshing = state.data == Data.Loading,
+        onRefresh = { viewModel.fetchData() },
+    )
 
     ReviewersScreenStateless(
         pullRequestNumber = state.pullRequestNumber,
         data = state.data,
+        pullRefreshState = swipeRefreshState,
```
jacek-marchwicki: suggestion: maybe we will stick to `pullRefreshState` name instead of `swipeRefreshState` what do you think?

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -98,25 +110,28 @@ private fun PullRequestsScreenStateless(
...
 @Composable
 private fun PullRequestContent(
-    state: Data.Success,
+    state: Data,
+    pullRefreshState: PullRefreshState,
     padding: PaddingValues,
     onAction: (PulLRequestsAction) -> Unit,
 ) {
-    if (state.pullRequests.isEmpty()) {
+    if ((state as Data.Success).pullRequests.isEmpty()) {
```
jacek-marchwicki: bug: there will be a crash, when state won't be Data.Success.

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -270,6 +302,10 @@ fun PullRequestsScreenErrorPreview() {
         PullRequestsScreenStateless(
             PullRequestState(Data.Error),
             onAction = {},
+            pullRefreshState = rememberPullRefreshState(
```
jacek-marchwicki: suggestion: Maybe it would be nice having a preview of the loading state?

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
```diff
@@ -151,20 +168,25 @@ private fun ReviewersScreenContent(
...
-    LazyColumn(
-        modifier = modifier.fillMaxWidth(),
-    ) {
-        itemsIndexed(reviewers) { index, reviewer ->
-            ReviewerItem(
-                reviewer = reviewer,
-                index = index,
-                onNotifyClick = onNotifyClick,
-            )
+    Box(modifier = Modifier.pullRefresh(pullRefreshState)) {
```
jacek-marchwicki: suggestion: This Box with PullRefreshIndicator is done twice for two different screens, maybe it's worth considereing as common pattern and extracting it as a component?

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
```diff
@@ -151,20 +168,25 @@ private fun ReviewersScreenContent(
...
+        LazyColumn(
+            modifier = modifier.fillMaxWidth(),
```
jacek-marchwicki: chore: I think the common practice is to pass `modifer` parameter from a function, should be passed to the first component (in this case to Box compoment). Not to the inner component (in this case LazyColumn component).

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
```diff
@@ -126,23 +140,26 @@ private fun ReviewersScreenStateless(
...
 @Composable
 private fun ReviewersScreenContent(
-    data: Data.Success,
+    data: Data,
+    pullRefreshState: PullRefreshState,
     padding: PaddingValues,
     onAction: (ReviewersAction) -> Unit,
 ) {
-    if (data.reviewers.isNotEmpty()) {
+    if ((data as Data.Success).reviewers.isNotEmpty()) {
```
jacek-marchwicki: bug: the same as above

jacek-marchwicki (changes requested):

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
Krzysiudan: **Quesion:** Do we want to pull refresh to be possible when the list is empty like there are no reviewers? If so we need to add it to he `EmptyListPlaceholder()` too.

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
Krzysiudan, line 1: **Nitpick** I noticed that on PullRequestScreen I can pull to refresh on the whole screen but on RewiewersScreen I can do it only from the list area (Only there where we have items).

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -126,17 +141,22 @@ private fun PullRequestsList(
...
-    LazyColumn(
-        modifier = modifier.fillMaxSize(),
-    ) {
-        itemsIndexed(pullRequests) { index, item ->
-            PullRequestItem(
-                index = index,
-                data = item,
-                onClick = onItemClick,
-            )
+    Box(modifier = Modifier.pullRefresh(pullRefreshState)) {
```
Krzysiudan: **Praise:** I haven't seen the implementation of pull refresh in compose, it seems simple. So glad to see that 👌

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsViewModel.kt
nowakweronika: No, it's not. When user doesn't pull to refresh, the old data is still presented.

nowakweronika: I don't think this test is necessary because the `WHEN init THEN display pull requests list` test is doing this job.

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersViewModel.kt
nowakweronika: No, it's not. When user doesn't pull to refresh, the old data is still presented.
I don't think this test is necessary because the `WHEN init THEN display pull requests list` test is doing this job.

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsViewModel.kt
jacek-marchwicki: > No, it's not. When user doesn't pull to refresh, the old data is still presented.

I'm talking about this full-screen loader, when doing pull-to-refresh:
full-screen-loader.webm

nowakweronika: Ok, now I understand :-)

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
nowakweronika: It's really hard to find the proper solution. I tried to add `.verticalScroll(rememberScrollState())` in many ways, but it always up the `EmptyListPlaceholder` to the top of the screen. Do you know how to do it without such a result?
Screenshot 2023-05-29 at 16 08 49

app/src/main/java/com/appunite/loudius/ui/components/LoudiusPullToRefreshBox.kt
```diff
@@ -0,0 +1,41 @@
...
+import androidx.compose.material.pullrefresh.PullRefreshIndicator
+import androidx.compose.material.pullrefresh.PullRefreshState
+import androidx.compose.material.pullrefresh.pullRefresh
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+
+@Composable
+fun LoudiusPullToRefreshBox(
+    content: @Composable () -> Unit,
```
Krzysiudan: **Suggestion:** You can move the lambda to the last parameter for easier use

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsViewModel.kt
```diff
@@ -58,10 +60,26 @@ class PullRequestsViewModel @Inject constructor(
     var state: PullRequestState by mutableStateOf(PullRequestState())
         private set
 
+    private val _isRefreshing = MutableStateFlow(false)
+    val isRefreshing = _isRefreshing.asStateFlow()
```
Krzysiudan, line 64: **Question:** What are the advantages of having `isRefreshing` as a separate state flow instead of having `isRefreshing` in `PullRequestState`?

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
Krzysiudan: Unfortunately, I don't know how to do it either. I tried but I failed. Maybe having that at the top is not that bad after all. 😅

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsViewModel.kt
nowakweronika: Actually, there are a few benefits😅
1. When we use `PullRequestState`, we want to use proper UI (ErrorScreen, LoadingIndicator, PRContent), in this case we have only edited PRContent, not added another screen.
2. In `when` statement we don't have to put the next option witch would be `isRefreshing` and thanks to that the code is more consistent in my opinion, because as I said in the first sentence, we don't have another screen for that.
3. Whole implementation is much easier and thus code is more readable : -)

Krzysiudan (approved):
Fine work, LGTM! 👌

nowakweronika:
@jacek-marchwicki

app/src/main/java/com/appunite/loudius/ui/components/LoudiusPullToRefreshBox.kt
```diff
@@ -0,0 +1,41 @@
...
+import androidx.compose.material.ExperimentalMaterialApi
+import androidx.compose.material.pullrefresh.PullRefreshIndicator
+import androidx.compose.material.pullrefresh.PullRefreshState
+import androidx.compose.material.pullrefresh.pullRefresh
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+
+@Composable
+fun LoudiusPullToRefreshBox(
```
jacek-marchwicki, line 34: suggestion: maybe it would be worth having a preview for this component?

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsViewModel.kt
jacek-marchwicki: suggestion,out-of-scope:
The current implementation is good enough. We might also consider having only one state across the app instead of two states. But I don't have strong opinions about netither of those two.

I think about, changing:
```kotlin
data class PullRequestState(
    val data: Data = Data.Loading,
    val navigateToReviewers: NavigationPayload? = null
)
```

to

```kotlin
data class PullRequestState(
    val data: Data = Data.Loading,
    val navigateToReviewers: NavigationPayload? = null,
    val isRefreshing: Boolean = false,
)
```

Also the code bellow becomes:
```kotlin
    fun refreshData() {
        viewModelScope.launch {
            state = state.copy(isRefreshing = true)
            pullRequestsRepository.getCurrentUserPullRequests()
                .onSuccess {
                    state = state.copy(data = Data.Success(it.items), isRefreshing = false)
                }.onFailure {
                    state = state.copy(data = Data.Error, isRefreshing = false)
                }
        }
    }
```

But you can keep the code as it is.

app/src/main/java/com/appunite/loudius/ui/components/LoudiusPullToRefreshBox.kt
```diff
@@ -0,0 +1,41 @@
...
+    pullRefreshState: PullRefreshState,
+    refreshing: Boolean,
```
jacek-marchwicki, line 36: nitpick: I don't really like passing `refreshing` state twice to this state. I mean about:
```kotlin
LoudiusPullToRefreshBox( 
    content = content,
    refreshing = false, // <--
    pullRefreshState = rememberPullRefreshState(
                refreshing = false, // <--
                onRefresh = {},
            ),
)
```

It's fine, but I'd love having this a little bit simpler ;) Passing the same state twice, is always a place for possible issues.

I wonder if this is possible to just pass:
```kotlin
LoudiusPullToRefreshBox( 
    content = content,
    refreshing = false,
    onRefresh = {},
)
```

or:

```kotlin
LoudiusPullToRefreshBox( 
    content = content,
    pullRefreshState = rememberPullRefreshState(
                refreshing = false,
                onRefresh = {},
            ),
)
```

app/src/test/java/com/appunite/loudius/ui/pullrequests/PullRequestsViewModelTest.kt
```diff
@@ -40,6 +40,17 @@ class PullRequestsViewModelTest {
     private val pullRequestRepository = spyk(FakePullRequestRepository())
     private fun createViewModel() = PullRequestsViewModel(pullRequestRepository)
 
+    @Test
```
jacek-marchwicki, line 45: suggestion: 
1. we might also have some test for ReviewersViewModel.
2. we might also have a test for `isRefreshing` field.

jacek-marchwicki (approved):
lgtm

app/src/main/java/com/appunite/loudius/ui/components/LoudiusPullToRefreshBox.kt
nowakweronika: I like it, done🤓

app/src/test/java/com/appunite/loudius/ui/pullrequests/PullRequestsViewModelTest.kt
nowakweronika: done :)

PR #97, opened 2023-05-25
[LD-98] Add github action for recording snapshots. [New snapshots]

Krzysiudan:
## Why? 
Generating snapshots on different OS sometimes results in a few different pixels. Generating ones on CI resolves that issue. 

## Changes
- added generating screenshots on CI by adding [New snapshots] to the pull request title. Otherwise, snapshots are verified on every pull request. 
- one new screenshot recorded (different rendering on Ubuntu then on mac-os for this particular snapshot)
- excluded `workflow` directory from jscpd linter checks (copy-paste checks) 

## Behaviour
Adding [New snapshots] to the pull request title results in recording new snapshots and auto-committing new ones to the repo. **Adds only the one that has changed with the commit**.
Pull requests without [New snapshots] in a title will run the `verify snapshots` job. 
Recording and verifying tasks show its result in a single comment.

### DIFF
--- a/.github/workflows/run-snapshot-generation.yml
+++ b/.github/workflows/run-snapshot-generation.yml
@@ -1,3 +1,4 @@
+---
 name: Snapshot recording
 
 on:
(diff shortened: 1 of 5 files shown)


Krzysiudan:
Record snapshot!

kezc (approved):
LGTM

.github/workflows/run-snapshot-generation.yml
```diff
@@ -0,0 +1,103 @@
+---
```
kezc, line 1: thought: maybe we could include in readme the information that if you add `[New snapshots]` the screenshots are recorded 🤔

.github/workflows/run-snapshot-generation.yml
```diff
@@ -0,0 +1,103 @@
...
+  generate_snapshots:
+    if: ${{contains(github.event.pull_request.title, '[New snapshots]')}}
+    name: Generate snapshots
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          lfs: true
+
+      - name: Set up Git LFS
```
jacek-marchwicki: question/suggestion: do we need  `Set up Git LFS`  if 
```kotlin
- uses: actions/checkout@v3
        with:
          lfs: true
```
is placed?

.github/workflows/run-snapshot-generation.yml
```diff
@@ -0,0 +1,103 @@
...
+          commit_message: "[Paparazzi] Record new snapshots"
+
+      - name: Upload snapshot record report
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: snapshot-recording-failure-report
+          path: /home/runner/work/Loudius/Loudius/components/build/reports/tests/testDebugUnitTest/
+
+      - name: LFS-warning
```
jacek-marchwicki: question: do we need this if it is already placed in https://github.com/appunite/Loudius/blob/feature/LD-91/golden-tests/.github/workflows/run-snapshot-test.yml#L25?

IF we need this shouldn't be `if: always()`?

.github/workflows/run-snapshot-generation.yml
```diff
@@ -0,0 +1,103 @@
...
+        with:
+          lfs: true
+
+      - name: Set up Git LFS
+        run: |
+          git lfs install
+
+      - name: Gradle - Record snapshots with Paparazzi
+        id: testStep
+        run: ./gradlew clean components:recordPaparazziDebug
```
jacek-marchwicki, line 30: question: upcoming tests has added `if: always()`, do we need this? I think the gradle command should succeed if new screenshots are recorded correctly?

jacek-marchwicki (approved):
lgtm

.github/workflows/run-snapshot-generation.yml
Krzysiudan: Thanks for that. Of course, you are right 👌

.github/workflows/run-snapshot-generation.yml
Krzysiudan: Yeah, you are right. I am wondering if we need that in the `run-snapshot-test.yml` too because now everything is done on CI. Maybe for now I will remove it only from `run-snapshot-generation.yml`.

.github/workflows/run-snapshot-generation.yml
Krzysiudan: Hmm. I think I don't understand. What do you mean by upcoming tests? Do you mean the steps after recording - with comment updating?

nowakweronika (approved):
Question: Why is snapshot verification skipped?

.github/workflows/run-snapshot-generation.yml
jacek-marchwicki: Indeed, I thought about "steps". I think `Gradle - Record snapshots with Paparazzi` step should always succeed. Even if new screenshots are generated. So if this step always succeeds, we don't need to add `if: failure()` and `if: always()` to steps that follow `Gradle - Record snapshots with Paparazzi` step. Am I right?

.github/workflows/run-snapshot-test.yml
```diff
@@ -1,3 +1,4 @@
+---
```
jacek-marchwicki, line 1: Why this is added?

.github/workflows/run-snapshot-generation.yml
jacek-marchwicki: Oh... Now I understand... because only one of those steps are executed at the same time, this is why you've added LFS-warning to both... this make sense in case of someone added files by them selfs.

.github/workflows/run-snapshot-test.yml
Krzysiudan: One of the linters has a warning that "File is not started with ---" , so I assumed that it is a standard way for .yaml files.

.github/workflows/run-snapshot-generation.yml
Krzysiudan: Actually it is possible that recording fails. For example, upgrading the ktx version to 1.10.0 make recording fail. That's a paparazzi bug. So if one bug exists and can fail the job then I believe it is helpful to have that.

.github/workflows/run-snapshot-generation.yml
Krzysiudan: Done that 👌

Krzysiudan:
> Question: Why is snapshot verification skipped?
That's because we have [New snapshots] added to the pull request title. Only one job is fired to the pull request - recording or verifying.

.github/workflows/run-snapshot-generation.yml
Krzysiudan: Ahh, in a readme. I thought about adding a comment in a yaml file. In readme it is even a better idea. 👌

PR #102, opened 2023-05-31
[LD-106] Screenshot when test fail

nowakweronika:
## Description
*We want to capture screenshot on firebase when test fail.*

## Documentation
Based on:
1. [Take a screenshot on Android Espresso Test Failure](https://andresand.medium.com/take-a-screenshot-on-android-espresso-test-failure-dbb7d06dfba6)
2. [The Firebase TestLab Intrumentation Library](https://github.com/firebase/firebase-testlab-instr-lib)

## Checklist
- [ ] You can add some wrong tests, commit them and check if screenshot will be captured :)


app/src/androidTest/java/com/appunite/loudius/util/EspressoScreenshot.kt
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Used to automatically capture screenshots of failed tests.
+ */
+object EspressoScreenshot {
+    private val imageCounter = AtomicInteger(0)
+    private val dotPNG = ".png"
+    private val underscore = "_"
+
+    // Firebase Test Lab requires screenshots to be saved to /sdcard/screenshots
+    // https://github.com/firebase/firebase-testlab-instr-lib/blob/f0a21a526499f051ac5074dc382cf79e237d2f4e/firebase-testlab-instr-lib/testlab-instr-lib/src/main/java/com/google/firebase/testlab/screenshot/FirebaseScreenCaptureProcessor.java#L36
```
jacek-marchwicki: question: If you find this library, don't you think it's good idea to use it? :)

There is readme:
https://github.com/firebase/firebase-testlab-instr-lib/blob/master/firebase-testlab-instr-lib/README.md

example usage: https://github.com/firebase/firebase-testlab-instr-lib/blob/master/notepad/app/src/androidTest/java/com/example/android/notepad/NotePadTest.java#L61

dependency: https://github.com/firebase/firebase-testlab-instr-lib/blob/master/notepad/app/build.gradle#L27

manifest change:
https://github.com/firebase/firebase-testlab-instr-lib/blob/master/notepad/app/src/androidTest/AndroidManifest.xml#L8-L14

app/src/main/AndroidManifest.xml
```diff
@@ -5,6 +5,8 @@
     <!--  Internet Permissions  -->
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"
```
jacek-marchwicki: suggestion: if this is only required for running tests, I'd prefer moving this to `app/src/debug/AndroidManifest.xml` file, so we won't add this permission to the rest of the app. 

question: maybe we can move this to `app/src/androidTest/AndroidManifest.xml` file, but I don't if this is going to work?

app/src/androidTest/java/com/appunite/loudius/util/ScreenshotTestRule.kt
```diff
@@ -0,0 +1,75 @@
...
+import androidx.test.espresso.Espresso
+import androidx.test.espresso.base.DefaultFailureHandler
+import androidx.test.platform.app.InstrumentationRegistry.getInstrumentation
+import org.junit.rules.TestRule
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * TestRule used to run all test methods try count 1 time. Take screenshots on failure.
```
jacek-marchwicki: question: I don't understand why we need this retry count?

nowakweronika: Thanks to that we can set how many times we want to start test and take screenshot.

app/src/debug/AndroidManifest.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools">
```
nowakweronika, line 3: AS said It's never used :)

app/src/androidTest/java/com/appunite/loudius/util/ScreenshotTestRule.kt
Krzysiudan, line 1: **Suggestion** Probably some git history cleaning would be valuable.

Krzysiudan (approved):
LGTM!

app/src/androidTest/java/com/appunite/loudius/util/ScreenshotTestRule.kt
kezc: The count variable has a private modifier, so it means that we'll have to modify the rule code to change. We could set it as public, however, I think this could be probably just another rule to run the test multiple times.

app/src/androidTest/java/com/appunite/loudius/util/ScreenshotTestRule.kt
```diff
@@ -0,0 +1,76 @@
...
+                for (i in 0 until tryCount) {
+                    errorHandled.set(false)
+                    try {
+                        base.evaluate()
+                        return
+                    } catch (t: Throwable) {
+                        if (!errorHandled.get()) {
+                            Screenshot.capture().process()
+                        }
+                        error = t
```
kezc: question: isn't that a problem that we only show the last stored exception? Maybe we could store them in a list, but then the complexity grows as you can just throw multiple exceptions, so you would need to use `MultipleFailuresError`.

app/src/androidTest/java/com/appunite/loudius/util/ScreenshotTestRule.kt
nowakweronika: I think you're right, I deleted it :)

PR #105, opened 2023-06-05
[LD-100] Fix UI bugs related to snapshots [New snapshots]

Krzysiudan:
## Changes:
- Removed external padding from the `LoudiusOutlinedButton` component
- Corrected text in the Loudius text placeholder

### DIFF
--- a/app/src/main/java/com/appunite/loudius/ui/login/LoginScreen.kt
+++ b/app/src/main/java/com/appunite/loudius/ui/login/LoginScreen.kt
@@ -23,6 +23,7 @@ import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/appunite/loudius/ui/login/LoginScreen.kt
```diff
@@ -81,7 +83,9 @@ fun LoginScreenStateless(
     ) {
         LoginImage()
         LoudiusOutlinedButton(
-            modifier = Modifier.fillMaxWidth(),
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(horizontal = 46.dp),
```
nowakweronika, line 88: Question: In the previous implementation, padding was added to `LoudiusOutlinedButton` if `style` was `LoudiusOutlinedButtonStyle.Large`. We don't have this condition here, is it ok?

kezc: We don't use Large Buttons anywhere besides previews, so the extra margin was probably omitted.

kezc (approved):

app/src/main/java/com/appunite/loudius/ui/login/LoginScreen.kt
Krzysiudan: Hmm, I removed the padding from the `LoudiusOutlinedButton` cause it had external padding. Such components shouldn't have a margin added to them. So it was added by mistake. 
And here in `LoginScreen` the margin is added so that the UI stays the same. 
I don't know what you mean about the condition. I believe it is the same condition, we have a large button, and we have the same padding as before, but added here. 🤔

Krzysiudan: This bug/mistake came up in one of my previous PR, but I've decided to fix it later. I could have said it before. 👍

PR #106, opened 2023-06-05
[LD-91] Snapshot tests for whole UI [New snapshots]

Krzysiudan:
## Why? 
To check how implemented snapshot flow will work with the whole project. 

Changes: 
- moved almost all files into the `components` module

## Note

- Do we want to have these changes on develop branch?


kezc (commented):
I really like what you've created here!!! one thing - I'm unsure if we should merge it as basically, everything is the `components` package.  We need to think if we want to modularize our app or not.

Krzysiudan:
> I really like what you've created here!!! one thing - I'm unsure if we should merge it as basically, everything is the `components` package. We need to think if we want to modularize our app or not.

Yeah, I like it too. The whole app covered in screenshots sounds awesome. But it comes with a price, basically, everything is in `components` module, so definitely there is a place to wonder how to do it if so.

PR #110, opened 2023-06-19
[LD-108] Ensure orphan Github Actions are cancelled, when new pull-request is pushed

Krzysiudan:
## Why? 
If GitHub Actions are not cancelled, this increases the costs of CI. If we ensure those that are orphans are cancelled, we will decrease CI costs.
[[LD-108] ](https://loudius.atlassian.net/jira/software/projects/LD/boards/3/backlog?epics=visible&selectedIssue=LD-108)

## Changes: 
- Added concurrency rule for run-ui-test workflow
- Removed cancel workflow action from prepare-android-env action since it sometimes didn't work (see image below). Used concurrency instead. 

image

## Note: 
Tests on Firebase won't be cancelled. Due to implementing that is very time-consuming it wasn't finished. #107  I've created a separate task for that in the backlog so that we have it in mind. [LD-110](https://loudius.atlassian.net/browse/LD-110)

## Test
See previous actions. All from previous run are cancelled. 
image

### DIFF
--- a/.github/actions/prepare-android-env/action.yml
+++ b/.github/actions/prepare-android-env/action.yml
@@ -3,11 +3,6 @@ description: "Cancels previous runs, set ups jdk, validates gradle wrapper and u
 runs:
   using: "composite"
   steps:
-    - name: Cancel previous runs for the same branch
-      uses: styfle/cancel-workflow-action@0.9.1
-      with:
-        access_token: ${{ github.token }}
-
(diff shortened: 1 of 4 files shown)


jacek-marchwicki (approved):
lgtm

nowakweronika (approved):

kezc (approved):

PR #111, opened 2023-06-20
[LD-105] Add additional workflow for notifying on slack about workflow failure.

Krzysiudan:
## Why? 
If we implement a slack notification, we won’t forget about failures after change is merged. [LD-105
](https://loudius.atlassian.net/jira/software/projects/LD/boards/3?selectedIssue=LD-105)

## Changes: 
- Added one additional workflow which notifies Slack if any of the workflows fails.  For now, it will be launched on every workflow fail so I could test it (The following workflow is not working till it is merged on develop).

### DIFF
--- a/.github/workflows/check-for-workflow-failure.yml
+++ b/.github/workflows/check-for-workflow-failure.yml
@@ -1,0 +1,22 @@
+name: Check for master or prod failure
+on:
+  workflow_run:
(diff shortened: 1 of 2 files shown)


kezc (approved):
you could've used pepe as emoji but good job anyway!

PR #114, opened 2023-06-23
[LD-105] Slack notification as a step in every workflow

Krzysiudan:
## Why?

If we implement a Slack notification, we won’t forget about failures after the change is merged. [LD-105](https://loudius.atlassian.net/jira/software/projects/LD/boards/3?selectedIssue=LD-105)

## Changes

- added additional step with notifying on Slach when failure on develop or master branch into every existing workflow file
- added missing concurrency in `snapshot recording` workflow. 

## Test
See our spammed channel - `loudius-internal`
Screenshot 2023-06-26 at 09 15 40


.github/workflows/run-snapshot-generation.yml
```diff
@@ -8,6 +8,10 @@ on:
   pull_request:
     types: [opened, edited, synchronize]
 
+concurrency:
```
kezc, line 11: 👍

kezc (approved):
LGTM

PR #115, opened 2023-06-23
[LD-105] Delete check-for-workflow-failure.yml

Krzysiudan:
## Why? 
Because it's not working as expected. It shows not relevant information in a notification as only the same commit and not the correct workflow name.  
Screenshot 2023-06-23 at 12 51 38

**Slack notification action would be moved into the existing workflow file.**


jacek-marchwicki (approved):
lgtm

PR #117, opened 2023-07-12
[LD-109] Add Showkase component browser. [New snapshots]

Krzysiudan:
## Why? 
To make an example implementation of [Showkase](https://github.com/airbnb/Showkase) library in a project and to show its capabilities. 

## Changes: 
- Added icon on the Login screen (top right corner) which moves to the Showkase component browser activity
- Grouped composable previews for easier browsing
- Mark previews which show whole screens with skip parameter set to true
- Recorded new screenshots (due to preview organizing a lot of screenshot names changed)

## How to test it

- [x] Click on the icon at the top
- [x] See if component browser works
image
Screenshot 2023-07-12 at 14 25 48

Screenshot 2023-07-12 at 14 26 21

### DIFF
--- a/app/src/main/java/com/appunite/loudius/ShowkaseRoot.kt
+++ b/app/src/main/java/com/appunite/loudius/ShowkaseRoot.kt
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2023 AppUnite S.A.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.appunite.loudius
+
+import com.airbnb.android.showkase.annotation.ShowkaseRoot
+import com.airbnb.android.showkase.annotation.ShowkaseRootModule
+
+@ShowkaseRoot
+class ShowkaseRoot : ShowkaseRootModule
--- a/app/src/main/java/com/appunite/loudius/ui/login/LoginScreen.kt
+++ b/app/src/main/java/com/appunite/loudius/ui/login/LoginScreen.kt
@@ -38,6 +38,7 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import androidx.hilt.navigation.compose.hiltViewModel
+import com.airbnb.android.showkase.annotation.ShowkaseComposable
 import com.airbnb.android.showkase.models.Showkase
 import com.appunite.loudius.BuildConfig
 import com.appunite.loudius.R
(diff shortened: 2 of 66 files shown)


app/src/main/java/com/appunite/loudius/ShowkaseRoot.kt
Krzysiudan, line 1: **Info:**  Showkase root is now in 2 modules. 
I needed to move the root to the app module so that in Showkase component browser composables from the `app` module be visible. Showkase objects in the app module can take components from `components` module, but the Paparazzi tests can't use the Showkase object from the app module (circular dependency).  That's why now there are 2 Showkase roots.

app/src/main/java/com/appunite/loudius/ui/login/LoginScreen.kt
```diff
@@ -149,8 +183,9 @@ fun LoginScreenPreview() {
     }
 }
 
-@Preview(showSystemUi = true, showBackground = true)
 @Composable
+@Preview(showSystemUi = true, showBackground = true, group = "Dialogs")
+@ShowkaseComposable(skip = true)
```
Krzysiudan, line 191: **Info:** Showkase has an [issue](https://github.com/airbnb/Showkase/issues/235) showing dialogs in the component browser. 
That's why I've skipped all the dialogs in the browser.

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -284,6 +287,7 @@ fun PullRequestsScreenEmptyListPreview() {
 }
 
 @Preview("Pull requests - Loading")
+@ShowkaseComposable(skip = true)
 @Composable
```
nowakweronika: Nitpick: If you will change something in codebase, you can change the order here😅
```
@Preview("Pull requests - Loading")
@Composable
@ShowkaseComposable(skip = true)
```

app/src/main/java/com/appunite/loudius/ui/pullrequests/PullRequestsScreen.kt
```diff
@@ -272,6 +274,7 @@ fun PullRequestsScreenPreview() {
 
 @Preview("Pull requests - empty list")
 @Composable
+@ShowkaseComposable(skip = true)
```
nowakweronika, line 277: Question: Why don't we have `group` field in `@Preview` annotation here?

nowakweronika: And below in this file?

app/src/main/java/com/appunite/loudius/ui/reviewers/ReviewersScreen.kt
```diff
@@ -309,6 +310,7 @@ private val successData = listOf(
 
 @Preview
 @Composable
+@ShowkaseComposable(skip = true)
```
nowakweronika, line 313: Question: Why don't we have group field in @Preview annotation here and below?

trOnk12: how would you propose to group it here ? 🤔 I honestly do not see any valid group for this, let me know if you have some idea

nowakweronika: I think that it's `full screen` or something like `app screen`. I meant we can't see this screen in components at all.

trOnk12: hey @nowakweronika  I looked what is going on , and from what I see the Showkase is added only to the component module and this is also the reason why it is genereting only stuff from within that module, that is why you are not seeing stuff from the app module, let me know if that answers your question and we can merge this guy to the develop

nowakweronika: Ok, thanks!

nowakweronika (approved):

PR #118, opened 2023-07-14
[LD-24] Add access token encryption.

Krzysiudan:
## Why? 
To make storing access tokens safer. 

## Changes:
- added encryption for an access token with androidx security library.


app/src/test/java/com/appunite/loudius/domain/UserLocalDataSourceTest.kt
```diff
@@ -27,21 +26,25 @@ import strikt.assertions.isEmpty
 import strikt.assertions.isEqualTo
 
 class UserLocalDataSourceTest {
-    private val sharedPreferences = FakeSharedPreferences()
```
Krzysiudan, line 30: **info** I couldn't create fake implementation for EncryptedSharePrefrences so I used mocks.

PR #120, opened 2023-09-14
Setup ktor and auth service in ktor

nowakweronika:
## Description
*Add Ktor instead of Retrofit in `AuthService` and adjust `AuthDataSource` and `...Module` classes.*

### Why?

### DIFF
--- a/app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
+++ b/app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
@@ -61,13 +61,14 @@ object NetworkModule {
     @Provides
     @Singleton
     @AuthAPI
-    fun provideAuthKtor(
+    fun provideAuthHttpClient(
         gson: Gson,
         @AuthAPI baseUrl: String,
         loggingInterceptor: HttpLoggingInterceptor,
     ): HttpClient = HttpClient(OkHttp) {
         expectSuccess = true
         engine {
+            addInterceptor(TestInterceptor)
             addInterceptor(loggingInterceptor)
         }
         defaultRequest {
--- a/app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
+++ b/app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
@@ -28,6 +28,12 @@ import dagger.Module
 import dagger.Provides
 import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
+import io.ktor.client.HttpClient
+import io.ktor.client.engine.okhttp.OkHttp
+import io.ktor.client.plugins.contentnegotiation.ContentNegotiation
+import io.ktor.client.plugins.defaultRequest
+import io.ktor.http.ContentType
+import io.ktor.serialization.gson.GsonConverter
 import okhttp3.OkHttpClient
 import okhttp3.logging.HttpLoggingInterceptor
 import retrofit2.Retrofit
@@ -55,21 +61,21 @@ object NetworkModule {
     @Provides
     @Singleton
     @AuthAPI
-    fun provideAuthRetrofit(
+    fun provideAuthKtor(
         gson: Gson,
         @AuthAPI baseUrl: String,
         loggingInterceptor: HttpLoggingInterceptor,
-    ): Retrofit {
-        val okHttpClient = OkHttpClient.Builder()
-            .addInterceptor(TestInterceptor)
-            .addInterceptor(loggingInterceptor)
-            .build()
-
-        return Retrofit.Builder()
-            .baseUrl(baseUrl)
-            .client(okHttpClient)
-            .addConverterFactory(GsonConverterFactory.create(gson))
-            .build()
+    ): HttpClient = HttpClient(OkHttp) {
+        expectSuccess = true
+        engine {
+            addInterceptor(loggingInterceptor)
+        }
+        defaultRequest {
+            url(baseUrl)
+        }
+        install(ContentNegotiation) {
+            register(ContentType.Application.Json, GsonConverter(gson))
+        }
     }
 
     @Provides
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
```diff
@@ -55,21 +61,21 @@ object NetworkModule {
     @Provides
     @Singleton
     @AuthAPI
-    fun provideAuthRetrofit(
```
trOnk12, line 58: question: Is this really returning AuthKtor ? Wouldn't it be a better idea to rename it to something like provideHttpClient? It is not something specific to authentication only from what I understand ?

trOnk12 (approved):
nothing blocking ! :) Looks good to me

app/src/main/java/com/appunite/loudius/network/datasource/AuthDataSource.kt
```diff
@@ -47,16 +45,15 @@ class AuthNetworkDataSource @Inject constructor(
...
-                } else {
-                    Result.failure(response.mapErrorToException())
-                }
+        authService.getAccessToken(clientId, clientSecret, code).flatMap { response ->
+            if (response.accessToken != null) {
+                Result.success(response.accessToken)
+            } else {
+                Result.failure(response.mapErrorToException())
             }
+        }
```
trOnk12, line 57: I am still not sure if we realllly need this layer, we could just do it inside AuthService

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
```diff
@@ -55,21 +61,21 @@ object NetworkModule {
...
-
-        return Retrofit.Builder()
-            .baseUrl(baseUrl)
-            .client(okHttpClient)
-            .addConverterFactory(GsonConverterFactory.create(gson))
-            .build()
+    ): HttpClient = HttpClient(OkHttp) {
+        expectSuccess = true
+        engine {
+            addInterceptor(loggingInterceptor)
```
jacek-marchwicki, line 72: chore: `TestInterceptor` is not registered

app/src/main/java/com/appunite/loudius/network/datasource/AuthDataSource.kt
nowakweronika: Can you give me an example how to do it in `AuthService`? :) Because in `AuthDataSource` we are retrieving `AccessTokenResponse` which can contain error or access token and then return `AccessToken` (which is alias for string) and I am wondering how to do it in `AuthService` and still have `AccessTokenResponse`.

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
nowakweronika: Good idea with HttpClient instead of Ktor! I'll change that to `provideAuthHttpClient`. It's specific to authentication, hence the name. We have `BASE_API_URL` and `AUTH_API_URL` and we're using the second one only in authentication. In next PR you will see `provideBaseHttpClient`. :)

app/src/main/java/com/appunite/loudius/network/datasource/AuthDataSource.kt
trOnk12: I would just  do all the above inside AuthService, but once again, it i just a suggestion nothing blocking if you really see a value in such a split that we could benefit in the future, please go ahead :P It would mean that the authservice would return either success or a failure like AuthDataSource does

trOnk12 (approved):

jacek-marchwicki (approved):
lgtm

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
```diff
@@ -55,21 +61,22 @@ object NetworkModule {
...
-            .addConverterFactory(GsonConverterFactory.create(gson))
-            .build()
+    ): HttpClient = HttpClient(OkHttp) {
+        expectSuccess = true
+        engine {
+            addInterceptor(TestInterceptor)
+            addInterceptor(loggingInterceptor)
+        }
+        defaultRequest {
+            url(baseUrl)
```
GhulJr, line 75: **Question:** Are you sure it works as intended? I remember issue like [this](https://youtrack.jetbrains.com/issue/KTOR-759), where paths of url did not join. 🤔 Maybe it has been solved, but just in case here is workaround 😉 
`url.takeFrom(URLBuilder().takeFrom(baseUrl).apply { encodedPath += url.encodedPath })`

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
```diff
@@ -55,21 +61,22 @@ object NetworkModule {
...
-            .client(okHttpClient)
-            .addConverterFactory(GsonConverterFactory.create(gson))
-            .build()
+    ): HttpClient = HttpClient(OkHttp) {
+        expectSuccess = true
+        engine {
+            addInterceptor(TestInterceptor)
+            addInterceptor(loggingInterceptor)
+        }
+        defaultRequest {
```
GhulJr, line 74: **Suggestion:** You can also specify default headers if they are needed, like this 😉 
```
contentType(ContentType.Application.Json)
accept(ContentType.Application.Json)
```

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
```diff
@@ -55,21 +61,22 @@ object NetworkModule {
...
-        return Retrofit.Builder()
-            .baseUrl(baseUrl)
-            .client(okHttpClient)
-            .addConverterFactory(GsonConverterFactory.create(gson))
-            .build()
+    ): HttpClient = HttpClient(OkHttp) {
+        expectSuccess = true
+        engine {
+            addInterceptor(TestInterceptor)
+            addInterceptor(loggingInterceptor)
```
GhulJr, line 72: **Suggestion:** Ktor has own logging plugin 😉 
```
install(Logging) {
        logger = Logger.ANDROID
        level = LogLevel.ALL
}
```

PR #121, opened 2023-09-15
chore: design system documentation

trOnk12:
# Why ? 

In order to make most out of the goodies we are having such as showkase and paparazi, we introduce the design system docs, that has very low maintance cost except of updating it whenever a new kind of components comes. 

It references to the screenshot generated by the paparazzi


components/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+
+# Dialogs
+
+Dialogs are a standard way to inform the user about some result, giving the user option to react to it.
+
+<details>
+    <summary>Code snippet</summary>
+    <br>
+https://github.com/appunite/Loudius/blob/8e616a9a05e668d11425a40c9f452eb3091dcc2a/components/src/main/java/com/appunite/loudius/components/components/LoudiusDialog.kt#L25-L48
```
jacek-marchwicki, line 8: suggestion: instead of `8e616a9a05e668d11425a40c9f452eb3091dcc2a` I'd prefer using `main` branch. Or even better, it would be if you can use a relative path. Then the URL will be always valid :)

components/README.md
```diff
@@ -0,0 +1,141 @@
...
+</details>
+
+### Standard
```
jacek-marchwicki: suggestion: currently those images takes a lot of space. Maybe it would be worth training to put them side-by-side by using a table widget?

components/README.md
```diff
@@ -0,0 +1,141 @@
+[TOC]
+
+
+
+# Dialogs
+
+Dialogs are a standard way to inform the user about some result, giving the user option to react to it.
+
+<details>
+    <summary>Code snippet</summary>
```
jacek-marchwicki, line 6: suggestion: I think it would be more beneficial for someone who read this documentation to read usage, then implementation of a widget. So you can link @Preview, rather then implementation.

components/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+#### Content only
+
+Component taking only the "content".
+
+
+
+<img src="src/test/snapshots/images/com.appunite.loudius_PaparazziShowkaseTests_preview_tests[Default Group-LoudiusListItemJustContentPreview,1,light].png" />
+
+# Buttons
```
nowakweronika, line 72: Question: What do you think about cut whole screen only to button?

nowakweronika (approved):
LGTM

components/README.md
nowakweronika: Ok, I read the description again and now I see why we have whole screens ;)

trOnk12: yea exactly I do not think there is a way around it ;s

trOnk12: I tried some approaches but in the end paparrazi will take a screenshot of the device so we will end up with abig screenshot, if you have some ideas let me know

components/README.md
trOnk12: good idea 🆗

components/README.md
nowakweronika: I was googling it for a while and unfortunately I couldn't find any helpful information :(

components/README.md
trOnk12: sure, good idea

components/README.md
trOnk12: I am going to leave as it is for now. Maybe something for future improvements

components/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+<details>
+    <summary>Code snippet</summary>
+    <br>
+https://github.com/appunite/Loudius/blob/8e616a9a05e668d11425a40c9f452eb3091dcc2a/components/src/main/java/com/appunite/loudius/components/components/LoudiusDialog.kt#L25-L48
+</details>
+| Variant                                                      | Snapshot                                                     |
+| ------------------------------------------------------------ | ------------------------------------------------------------ |
+| **Standard** <br />This component gives the user option to confirm the information. | <img src="src/test/snapshots/images/com.appunite.loudius_PaparazziShowkaseTests_preview_tests[Default Group-LoudiusDialogSimplePreview,1,light].png" alt="alt text" style="zoom: 80%;" /> |
+| **Advanced**<br />This component gives  the user the choice out of two options, usually dismiss or confirmation. | <img src="src/test/snapshots/images/com.appunite.loudius_PaparazziShowkaseTests_preview_tests[Default Group-LoudiusDialogAdvancedPreview,1,light].png" alt="alt text" style="zoom: 80%;" /> |
```
jacek-marchwicki: chore: the current solution doesn't work: https://github.com/appunite/Loudius/blob/chore/design_system_docs/components/README.md

jacek-marchwicki (changes requested):

PR #126, opened 2023-09-20
user api

nowakweronika:
## Description
*Add Ktor instead of Retrofit in `UserService` and adjust `UserDataSource` and `...Module` classes.*
### Why?

### DIFF
--- a/app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
+++ b/app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
@@ -82,6 +82,31 @@ object NetworkModule {
     @Provides
     @Singleton
     @BaseAPI
+    fun provideBaseHttpClient(
+        gson: Gson,
+        @BaseAPI baseUrl: String,
+        loggingInterceptor: HttpLoggingInterceptor,
+        authInterceptor: AuthInterceptor,
+        authFailureHandler: AuthFailureHandler,
+    ): HttpClient = HttpClient(OkHttp) {
+        expectSuccess = true
+        engine {
+            addInterceptor(authInterceptor)
+            addInterceptor(TestInterceptor)
+            addInterceptor(AuthFailureInterceptor(authFailureHandler))
+            addInterceptor(loggingInterceptor)
+        }
+        defaultRequest {
+            url(baseUrl)
+        }
+        install(ContentNegotiation) {
+            register(ContentType.Application.Json, GsonConverter(gson))
+        }
+    }
+
+    @Provides
+    @Singleton
+    @BaseAPI
     fun provideBaseRetrofit(
         gson: Gson,
         @BaseAPI baseAPIUrl: String,
--- a/app/src/main/java/com/appunite/loudius/network/datasource/UserDataSource.kt
+++ b/app/src/main/java/com/appunite/loudius/network/datasource/UserDataSource.kt
@@ -18,7 +18,6 @@ package com.appunite.loudius.network.datasource
 
 import com.appunite.loudius.network.model.User
 import com.appunite.loudius.network.services.UserService
-import com.appunite.loudius.network.utils.ApiRequester
 import javax.inject.Inject
 import javax.inject.Singleton
 
@@ -28,11 +27,7 @@ interface UserDataSource {
 
 @Singleton
 class UserDataSourceImpl @Inject constructor(
-    private val userService: UserService,
-    private val apiRequester: ApiRequester,
-) :
-    UserDataSource {
-    override suspend fun getUser(): Result<User> = apiRequester.safeApiCall {
-        userService.getUser()
-    }
+    private val userService: UserService
+) : UserDataSource {
+    override suspend fun getUser(): Result<User> = userService.getUser()
 }
--- a/app/src/main/java/com/appunite/loudius/network/services/UserService.kt
+++ b/app/src/main/java/com/appunite/loudius/network/services/UserService.kt
@@ -17,11 +17,34 @@
 package com.appunite.loudius.network.services
 
 import com.appunite.loudius.network.model.User
+import io.ktor.client.HttpClient
+import io.ktor.client.call.body
+import io.ktor.client.request.get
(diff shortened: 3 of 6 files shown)


app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
```diff
@@ -79,6 +79,31 @@ object NetworkModule {
...
+        gson: Gson,
+        @BaseAPI baseUrl: String,
+        loggingInterceptor: HttpLoggingInterceptor,
+        authInterceptor: AuthInterceptor,
+        authFailureHandler: AuthFailureHandler,
+    ): HttpClient = HttpClient(OkHttp) {
+        expectSuccess = true
+        engine {
+            addInterceptor(authInterceptor)
+            addInterceptor(TestInterceptor)
```
trOnk12, line 95: question : do we always want to add TestInterceptor ? I am not a ktor expert but my six sense is telling my it may not be good idea ?

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
```diff
@@ -79,6 +79,31 @@ object NetworkModule {
...
+            addInterceptor(AuthFailureInterceptor(authFailureHandler))
```
trOnk12, line 96: question: why can't we provide the AuthFailureInterceptor but we do provide authFailureHandler ? 🤔

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
```diff
@@ -79,6 +79,31 @@ object NetworkModule {
...
 
+    @Provides
+    @Singleton
+    @BaseAPI
+    fun provideBaseHttpClient(
+        gson: Gson,
+        @BaseAPI baseUrl: String,
+        loggingInterceptor: HttpLoggingInterceptor,
+        authInterceptor: AuthInterceptor,
+        authFailureHandler: AuthFailureHandler,
```
trOnk12, line 90: question: not related to this PR but why do we need two seperate classes for AuthInterceptor and AuthFailureHandler? Can't we just do everything inside the AuthInterceptor ?

app/src/main/java/com/appunite/loudius/network/datasource/UserDataSource.kt
```diff
@@ -28,11 +27,7 @@ interface UserDataSource {
 
 @Singleton
 class UserDataSourceImpl @Inject constructor(
```
trOnk12, line 29: please do not get mad at me :P I do not want to push anything, but I just can't help but notice, once again in this example that UserDataSource is not really useful, it is just a wrapper not adding anything new but creating unnecessary jumping between classes ;p

app/src/main/java/com/appunite/loudius/network/services/UserService.kt
```diff
@@ -17,11 +17,26 @@
...
-    @Headers("Accept: application/json")
-    @GET("user")
-    suspend fun getUser(): User
+
+    suspend fun getUser(): Result<User>
+}
+
+class UserServiceImpl @Inject constructor(private val client: HttpClient) : UserService {
+
+    override suspend fun getUser(): Result<User> = runCatching {
```
trOnk12, line 35: nice use of runCatching 👍

trOnk12 (commented):
well done 🔥  I have some questions not really related to the stuff you are addressing in this PR, but maybe it is a good moment to go back to them and address them if necessary ? let me know what you think

app/src/main/java/com/appunite/loudius/network/datasource/UserDataSource.kt
nowakweronika: Honestly, I strongly agree in that case 😅

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
nowakweronika: As I see both implementations, I think they're doing different things. Can you give me an example implementation of `AuthInterceptor` with `AuthFailureHandler` inside?

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
nowakweronika: I think we can provide this :) What benefits will it give us? Also I'm not sure if `AuthFailureInterceptor` should be annotated with singleton or not.

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
nowakweronika: We want to provide `TestInterceptor` because of UI tests. You can see that in `MockWebServerRule` in 63 line we're providing new url.

trOnk12: I am too newbie to understand what exactly is happening there :P But this is just abit weird that we are providing TestInterceptor to something that may be used in production , are we able to provide a httpClient specfic only to tests ? I feel like we are mixing here two responsibilites : that for tests and for production stuff.

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
trOnk12: in this case I guess there will not be a huge benefit, because we are creating the object just in that one place, but if in the future it will be reused somewhere or will be more complicated, the provideBaseRetrofit will grow bigger. 

I also do not see a reason or benefit not to do it, they are both annotated with @Inject so it should work out of the box anyways, no additonal work needed.

app/src/main/java/com/appunite/loudius/di/NetworkModule.kt
trOnk12: forget it from I read the code abit more and it looks like AuthFailureHandler has authFailureFlow, which needs to be Injectable as standalone, so I guess there is no way around it.

trOnk12 (approved):

PR #127, opened 2023-09-21
Pull requests api

nowakweronika:
## Description
*Add Ktor instead of Retrofit in `PullRequestsService` and adjust `PullRequestsDataSource` and `...Module` classes.*

### Why?


trOnk12 (approved):
🥇

PR #142, opened 2023-10-12
LD-124 Add robolectric tests support 

jacek-marchwicki:
### Why?
androidTests are very useful, they are executed on real devices and those tests ensures that app works on user devices. Unforuntelly those tests are slow to execute. Running the full test suite takes time, and requires having emulator.
There is an alternative for androidTests wich is Robolectric tests. Those tests uses the same espresso/androidx test framework, but are executed on local machine, not on device. Those tests are much faster.

| androidTest | robolectric test | unit tests |
|-|-|-|
| 👎🏾 very slow   | 👍🏾 faster   | 👍🏾 fastest |
| 👎🏾 expensive to run | 👍🏾 inexpensive | 👍🏾 inexpensive |
| 👍🏾 checks on real device | 👍🏾👎🏾simulate android environment | 👎🏾 run only on mocks |
| 👍🏾 runs the whole app on real device | 👍🏾👎🏾 runs the whole app on simulated device | 👎🏾 only checks part of the code | 
| 👎🏾 don't track the root of the problem | 👎🏾 don't track the root of the problem | 👍🏾 find the root of the problem | 
| 👎🏾 Flaky | 👎🏾👍🏾 a little bit flaky | 👍🏾 usually not flaky | 

What problems can be detected by given test suite
| Problem | androidTest | robolectric  | unit tests |
|-|-|-|-|
| Bitmap causes OutOfMemoryException |  ✅ | ❌ | ❌ |
| Issue in a certain android version |  ✅ | ❌ / ✅ | ❌ |
| There is a problem in app initialization |  ✅ | ✅ | ❌ |
| Problems related to other apps |  ✅ | ❌ | ❌ |
| Problems with communication to a backend |  ✅ | ✅ | ✅ |
| Activity initialization problem |  ✅ | ✅ | ❌ |
| Issue in android OS |  ✅ | ❌ | ❌ |
| Specific case/problem in a single class |  ❌ | ❌ | ✅ |
| Issues with progard | ❌* | ❌ | ❌ |

This leads to conclustions, that we have a piramide of test suite, that no of tests looks like this:
```mermaid
flowchart TD
    A[Unit Tests] -->|Less tests| B[Robolectric Tests] --> |Less tests| C[Android Tests] --> |Less tests| D[End To End Tests]
```

What is greatest is that Robolectric and Android Tests uses the same framework for testing which mean, we can write the same test once and run them once with Robolectric framework, once with Android Emulator.
I.e. for quick feedback, after developing a feature with a test. We can firstly ensure that given test runs in Robolectric environemnt with very quick feedback loop. Then run it on Android Emulator for a little bit slower feedback loop. Then Run the whole Robolectric test suite on our machine for quick feedback loop of the whole test suite. After this, we can push to CI to run the while test suite on Android Emulators to ensure if there is no regression on real devices.

## What?
* Added Robolectric framwork
* Set-up common space for Robolectric and Android Tests.
* Prepared set-up that allows running given test with Robolectric and Android Test mode.
* Migrated all Android Tests to Robolectric/Android tests
* Changed syncrhonization from `AndroidComposeTestRule.registerIdlingResource` which is not supported by robolectric to waitUntil syncrhonization [waitUntil](https://developer.android.com/reference/kotlin/androidx/compose/ui/test/junit4/ComposeTestRule#waitUntil(kotlin.Long,kotlin.Function0))
* Ensured all tests are executed on CI
* Created failed CI tests report available on GitHub https://github.com/appunite/Loudius/pull/142/checks?check_run_id=17675255175

## Links to related issues

- Fixes [LD-124](https://loudius.atlassian.net/browse/LD-124)

## Documentation
- Link to documentation:
  - [Robolectric - AndroidX tests](https://robolectric.org/androidx_test/)
  - [How to set robolectric tests](https://developer.android.com/training/testing/local-tests)
  - [Dcoumentation about jetpack compose testing](https://developer.android.com/jetpack/compose/testing)
  - [Robolectric - how to log?](https://stackoverflow.com/questions/10219915/where-is-log-output-written-to-when-using-robolectric-roboguice)
  - [How to set accessiblity for the app](https://developer.android.com/jetpack/compose/accessibility)
  - [Junit5 with junit4 support = junit5 Vintage](https://junit.org/junit5/docs/current/user-guide/#running-tests-ide-intellij-idea)
  - [How to reuse the same test across androidTests and unit tests](https://stackoverflow.com/questions/72218645/shared-srcdirs-between-test-and-androidtest-unresolved-references-after-upgrade)

## Notes
Robolectric doesn't support `AndroidComposeTestRule.registerIdlingResource` so we need to wait differently for loading to finish.

## Checklist

- [x] Functionality covered by unit tests
- [x] I have manually tested if the code works


jacek-marchwicki:
@nowakweronika

nowakweronika (approved):
LGTM

PR #159, opened 2023-10-24
[LD-126] Improve PR template and ensure it is set correctly

jacek-marchwicki:
### Why?

1. We've had PR template that doesn't match requiements of Loudius project
2. PR's are often missing some important parts in the description

### What?
1. Description: `Why?` is more important than `What?` when introducing changes
2. Description: `[LD-ZZZ](https://loudius.atlassian.net/jira/software/projects/LD/boards/3)` changed to `[LD-ZZZ](https://loudius.atlassian.net/browse/LD-ZZZ)` to link correct task
3. Description: `New functionality is covered by Analytics events` checklist item is inrelevant to Loudius, because we don't have analytics events
4. Description: Added `I have updated PR description with relevant information` checklist item to make sure noone forgot about PR description.
5. Description: Added `How to test` step, to ensure reviewer knows how to test given feature.
6. GitHub Action: Added Github Action that verifies if PR is described correctly.

## Links to related issues

- Fixes [LD-126](https://loudius.atlassian.net/browse/LD-126)

## Demo

Example failure:
Screenshot 2023-10-24 at 12 44 07
Then you see failed build with description why given CI failed:
Screenshot 2023-10-24 at 17 08 33

Here you can find example failure: https://github.com/appunite/Loudius/actions/runs/6628820563/job/18006755981?pr=161

You can check https://github.com/appunite/Loudius/pull/161 PR

## How to test

* **Step 1**: Create PR to this branch
  * Copy PR description from `pull_request_template.md` and set it.
* **Step 2**: Make fun updating PR description and ensure it fails when it should ;)
* **Expected:** PR should fail with good description when the title is incorrect.

**Note**: You can use for playground following PR: https://github.com/appunite/Loudius/pull/161 PR

## Documentation
- [Github Documentation](https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request)
- Keep in mind, that for new PRs template is updated after change is merged. So we need to wait for PR template update after this change is merged to `develop` branch.

## Checklist

~- [ ] Functionality covered by unit tests~
~- [ ] Functionality covered by integration tests~
- [x] I have updated PR description with relevant information
- [x] I have manually tested if the code and app works

PR #161, opened 2023-10-24
chore: test PR that should fail because of #159

jacek-marchwicki:


jacek-marchwicki:
Not needed anymore because #159 is already merged

PR #165, opened 2023-10-26
[LD-136] Firebase analytics

nowakweronika:
# Pull-Request
### Why?

We want to add analytics to project to track user behaviour.

### What?

Added analytics to project.

## Links to related issues

- Fixes [LD-136](https://loudius.atlassian.net/browse/LD-136), [LD-140](https://loudius.atlassian.net/browse/LD-140)

## Demo

**With analytics dashboard, you can see a bunch of informations!**
| Description | Picture |
|:---------------:|:--------------:|
| Users in last 30 minutes. | Screenshot 2023-10-30 at 15 06 42 |
| List of events from the selected period. | Screenshot 2023-10-30 at 15 06 52 |
| When you click on an event, you can see more detailed information. | Screenshot 2023-10-30 at 15 11 03 |

## How to test

First, you should add commands to terminal which allow you logging events.
```
$ adb shell setprop log.tag.FA VERBOSE
$ adb shell setprop log.tag.FA-SVC VERBOSE
$ adb logcat -v time -s FA FA-SVC
```
Last command will print logs for the device.
If you have problems with adb command, those links helped me: [Not found adb Error](https://dev.to/ravics09/solution-of-command-not-found-adb-error-29e7) and [StackOverflow question](https://stackoverflow.com/questions/50456824/adb-command-not-found-on-mac-computer).
### Then, you can start testing!
* **Step 1**: Open and log in to Loudius.
* **Step 2**: Click on one of the pull requests.
  * If you don't have any, create one and request reviewers.
* **Step 3**: Click on notify button.
* **Expected:** Now in Logcat you should be able to see relevant events. Please try to play with the internet connection and see the output 🤓

**❗️Be aware that you won't see analytics on dashboard immediately. Firebase needs about a few hours to note it.❗️**

## Documentation
- Link to [Firebase Analytics Docs](https://firebase.google.com/docs/analytics)
- Link to [Analytics Dashboard](https://console.firebase.google.com/u/0/project/loudius-fce48/analytics/app/android:com.appunite.loudius/overview/~2F%3Ft%3D1698317862287&fpn%3D999703798548&swu%3D1&sgu%3D1&params%3D_u..pageSize%253D25%2526_u.dateOption%253Dtoday&cs%3Dapp.m.dashboard.overview&g%3D1)
- Link to [Events Dashboard](https://console.firebase.google.com/u/0/project/loudius-fce48/analytics/app/android:com.appunite.loudius/events/~2Foverview%3Ft%3D1698319954101&fpn%3D999703798548&swu%3D1&sgu%3D1&params%3D_u..pageSize%253D25%2526_u.dateOption%253Dtoday%2526_r..layout.pageNumber%253D0&cs%3Dapp.m.events.overview&g%3D1)

## Checklist

~- [ ] Functionality is covered by unit tests~
~- [ ] Functionality is covered by integration tests~
- [x] I've updated PR description with relevant information
- [x] I've done self code review
- [x] I've manually tested if the code and app works


.github/workflows/run-test.yml
```diff
@@ -32,6 +32,14 @@ jobs:
         with:
           lfs: true
 
+      - name: Create file
```
jacek-marchwicki: **chore**: ❗❗❗❗ Never print secrets to console, because this is an open source project, and someone could read this file from GitHub Actions executions log. This might cause security leek of this key. 

Great that this file does not yet exist during the workflow execution. So it wasn't printed. But still remove this step!

.github/workflows/run-test.yml
```diff
@@ -64,6 +72,14 @@ jobs:
...
```
jacek-marchwicki: **chore**: ❗❗❗❗ Never print secrets to console, because this is an open source project, and someone could read this file from GitHub Actions executions log. This might cause security leek of this key. 

Great that this file does not yet exist during the workflow execution. So it wasn't printed. But still remove this step!

.github/workflows/run-test.yml
```diff
@@ -32,6 +32,14 @@ jobs:
...
+        run: cat /home/runner/work/Loudius/Loudius/app/google-services.json | base64
+
+      - name: Add google-services.json
+        env:
+          GOOGLE_SERVICES_JSON: ${{ secrets.GOOGLE_SERVICES_JSON }}
+        run: echo $GOOGLE_SERVICES_JSON > /home/runner/work/Loudius/Loudius/app/google-services.json
```
jacek-marchwicki: **chore**: Here absolute path is used. This might cause build issues if docker home path image is updated and our code would be checked-out to a different directory. The path also changes if GitHub project is renamed or forked.
It is safer to just use relative path, so instead of using `/home/runner/work/Loudius/Loudius/app/google-services.json`, just use `app/google-services.json`

.github/workflows/run-test.yml
```diff
@@ -64,6 +72,14 @@ jobs:
...
```
jacek-marchwicki: **chore**: Use relative path instead of absolute path.

.github/workflows/run-test.yml
```diff
@@ -64,6 +72,14 @@ jobs:
         with:
           lfs: true
 
+      - name: Create file
+        run: cat /home/runner/work/Loudius/Loudius/app/google-services.json | base64
+
+      - name: Add google-services.json
```
jacek-marchwicki, line 79: **question**: is it possible to build the app without the `google-services.json` file, or it is necessary to have one on your computer? 

I'm concern about the situation when someone forks this repo, and doesn't have access to this private file. Will that person be able to build the app locally and play with it?

app/src/main/java/com/appunite/loudius/analytics/events/ReviewersEvents.kt
```diff
@@ -0,0 +1,95 @@
...
+import com.appunite.loudius.analytics.StringEventParameter
+import com.appunite.loudius.analytics.util.EventsConstants.ITEM_NAME
+import com.appunite.loudius.analytics.util.EventsConstants.SUCCESS
+
+interface ReviewersEvent: Event
+
+object ClickNotifyEvent: ReviewersEvent {
+    override val name: String = "button_click"
+    override val parameters: List<EventParameter<*>> = listOf(
+        StringEventParameter(ITEM_NAME, "notify")
```
jacek-marchwicki: **suggestion**: I know that constants are common practice, but in case of tracking analytics events, those constants make harder to understand what actually is tracked.

I.e. when you read documentation for analytics events that usually looks like this:

| Parameter | Value | Type |
|-|-|-|
| event_name | button_click | string |
| item_name | notify | string |
| success | true/false | boolean |


it's obvious that:
```kotlin
object ClickNotifyEvent: ReviewersEvent {
    override val name: String = "button_click"
    override val parameters: List<EventParameter<*>> = listOf(
        StringEventParameter("item_name", "notify")
        BooleanEventParameter("success", true)
}
```
is correct, but if you have:
```kotlin
object ClickNotifyEvent: ReviewersEvent {
    override val name: String = "button_click"
    override val parameters: List<EventParameter<*>> = listOf(
        StringEventParameter(ITEM_NAME, "notify"),
        StringEventParameter(SUCCESS, true),
}
```
You need to check two places to ensure everything is fine. Also, when you implement new event, you need to create constants and jump between two files. This doesn't give much value. 
Also, if you have those consts, it is much easier to create type, than just copy-pasting values from a documentation.

app/src/main/java/com/appunite/loudius/analytics/EventParameter.kt
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.appunite.loudius.analytics
+
+sealed class EventParameter<T> {
```
jacek-marchwicki: **opinion**: when using sealed class, it is usually a little bit simpler to just use inner classes, especially if all are specified in a single file:

```kotlin
sealed class EventParameter<T> {
   abstract val name: String
   abstract val value: T
   data class String(
    override val name: String,
    override val value: String
) : EventParameter<String>()
data class Boolean(
    override val name: String,
    override val value: Boolean
) : EventParameter<Boolean>()
```

* This gives better autocompletion: If you Type `EventParameter.` than IDE will help you fill the rest, and present all possible autocomplitions.
* It gives you standard naming for sealed class. In this case you're naming class like: `StringEventParameter`, but it could be also: `EventParameterString`, or also `ParameterString`. In case of inner sealed classes you always have: `EventParameter.<Somting>`

**suggestion**:
You may consider skipping using generics, because `value` doesn't need to be in the base class. I mean something like this :
```kotlin
sealed class EventParameter {
    abstract val name: String
}

data class StringEventParameter(
    override val name: String,
    val value: String
) : EventParameter()

data class BooleanEventParameter(
    override val name: String,
    val value: Boolean
) : EventParameter()
```

You won't need to type `EventParameter<*>` everywhere :)

It's a little simple, but both solutions are correct :) 

All together:
```kotlin
sealed class EventParameter {
    abstract val name: String

    data class String(override val name: String, val value: String) : EventParameter()

    data class Boolean(override val name: String, val value: Boolean) : EventParameter()
}
```

jacek-marchwicki (changes requested):

.github/workflows/run-test.yml
nowakweronika: To build the app, you must have this file on your computer. Good point because before merging these changes, we should add the relevant information in the readme on how to create it and add it to the project. I will do it.

jacek-marchwicki: Is it possible, to make it optional, by some feature-flag? I.e. if the flag is missing, everything works, but analytics are not sent? I'm concern about a situation where someone outside AU downloads the project, and just simple want to play with it. And the person doesn't have the file.

nowakweronika: I think so but I have to make sure.

app/src/main/java/com/appunite/loudius/analytics/EventParameter.kt
nowakweronika: How simple!🤩

.github/workflows/run-test.yml
nowakweronika: Added in `app/build.gradle` 🥳

app/src/test/java/com/appunite/loudius/analytics/EventParametersConverterTest.kt
```diff
@@ -0,0 +1,55 @@
...
+        EventParameter.String(PARAM_NAME1, STRING_VALUE1),
+        EventParameter.String(PARAM_NAME2, STRING_VALUE2),
+        EventParameter.Boolean(PARAM_NAME3, BOOLEAN_VALUE)
+    )
+
+    private val converter = EventParametersConverter()
+
+    @Test
+    fun testConvert() {
+        val result = converter.convert(exampleParameters)
```
jacek-marchwicki: **nitpick**: I'd completely remove those constans (`PARAM_NAME1`, `exampleParameters`, etc,) :)

app/src/test/java/com/appunite/loudius/analytics/FirebaseAnalyticsEventTrackerTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import io.mockk.mockk
+import io.mockk.mockkStatic
+import io.mockk.verify
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@RunWith(RobolectricTestRunner::class)
```
jacek-marchwicki: **suggestion**/**question**: I think it doesn't need to be `RobolectricTestRunner` test. But even if it doesn't need its not a big deal

jacek-marchwicki (approved):
lgtm

